PR #26, opened 2016-12-29
Support for InfluxDB

dbhorst:
I added support for a custrom influx db to the firmware. You can then use grafana to display the data.


ricki-z:
Please use send_data and write a function to build a new data string for influxdb.

dbhorst:
The send_data methode uses SSL for ports != 80.
I have currently some problems with SSL in the influxdb 1.1 config. Until then, I can't optimize the code.
Secondly I do not know if influxdb is looking at the "Content-Type: application/json" field, because the content is not json.

ricki-z:
I've added your influxdb changes to my firmware. Please test the actual version. http client is now default. Content type of request can be set as parameter for sendData().
Custom api config is removed from WiFiConfig. We can only use 10 custom fields. Otherwise the firmware will crash on calling WiFiConfig.

PR #28, opened 2017-01-02
removed yield 

odon4t4:
due to of missing Scheduler https://www.arduino.cc/en/Reference/SchedulerYield


ricki-z:
yield() is needed. We use the Arduino IDE to program an ESP8266, not an Arduino.
ESP8266 needs the yield() to get time to hold the wifi connection (key refreshs,broadcast answers...). Otherwise connection will get lost after some time.
So merge will be refused.

odon4t4:
Thanks for the advise. This is the only way to learn something about arduinos libs.
Thanks

PR #30, opened 2017-01-10
Wi fi info

odon4t4:
I've added a First attempt for a info page over WiFi. 
I had tried to add some historical data and to plot them with SVG but the WiFiServer Lib can't handle to large html sites (mine is >6000 chars)

Also I suggest to eliminate the String class from the project to to the difficulties with dynamic mem allocation on embedded systems. (I also used the String class to increase the readability of the .ino)


ricki-z:
Can you send here the following information, arduino ide will print this after compilation:
used flash memory, should be below 50% for auto update
memory used by global vars, should be below 50.500 byte, otherwise firmware will crash on api connect

ricki-z:
I'm rewriting the wifi config part with some info pages, so i will merge your links into the appropriate pages.

PR #31, opened 2017-01-14
Added small webserver

user8465723:
- Anzeige des letzten Datensatzes


ricki-z:
An info page will be included in the next larger update. Together with a new WifiConfig. This needs some time to implement and testing.

PR #32, opened 2017-01-14
Adafruit MQTT encrypted

user8465723:
Adafruit IO is free and allows simple dashboards with graphs


ricki-z:
MQTT is rejected for our firmware. please look at issue #33 for reasons.

amotl:
You can use [luftdaten-to-mqtt] for converging data from the [luftdaten.info live api] to MQTT, see also #33 for more information.

[luftdaten-to-mqtt]: https://github.com/daq-tools/kotori/blob/master/kotori/vendor/luftdaten/luftdaten_api_to_mqtt.py
[luftdaten.info live api]: https://api.luftdaten.info/static/v1/data.json

PR #46, opened 2017-02-08
Add send_now variable to DRY up the code

corny:


ricki-z:
This won't compile (line 2055: starttimee not defined). And we have to save space on vars, so don't define unnecessary vars.

corny:
Opps, I fixed the error in line 2055. Instead of using a global variable this variable could also be local and passed to `sensorSDS()`.

corny:
Should we introduce a global `send_now` variable or a macro?

ricki-z:
I had changed the code so that your PR couldn't be merged automatically. But I've included your changes in the last version (global var send_now).

corny:
Thanks a lot!

PR #47, opened 2017-02-09
Refactor initialization of BME2180

corny:


ricki-z:
Please change sprintf to debug_out() in function. So output will be sent to the right interface.

ricki-z:
Please run compile before PR. The function name and the function call use different names, init_bme280() and initBME280().

corny:
sorry, thanks for fixing and merging

PR #51, opened 2017-02-12
Fix saving of configuration

corny:
Broken since ea1b214bde406a6158ff2b826dfd9b25b11ec5ed

The value of the submit button did not match the condition in the handler (`Save` != `Speichern`).
Checking the request method is better since there is only one form.
Even better would be routing based on the request method.

### DIFF
--- a/esp8266-arduino/ppd42ns-wificonfig-ppd-sds-dht/ppd42ns-wificonfig-ppd-sds-dht.ino
+++ b/esp8266-arduino/ppd42ns-wificonfig-ppd-sds-dht/ppd42ns-wificonfig-ppd-sds-dht.ino
@@ -805,7 +805,7 @@ void webserver_config() {
 	page_content.replace("{mac}",WiFi.macAddress());
 	page_content.replace("{fw}",SOFTWARE_VERSION);
 	page_content += FPSTR(WEB_CONFIG_SCRIPT);
-	if ((!server.hasArg("submit")) || (server.arg("submit") != "Save")) {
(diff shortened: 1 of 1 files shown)


corny:
If you really want to compare the button value, please put it in a macro or variable.
I also recommend to check the request method since a server should not modify the configuration on GET requests.

corny:
RFC 2616, [9.1.1 Safe Methods](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html):
> In particular, the convention has been established that the GET and HEAD methods SHOULD NOT have the significance of taking an action other than retrieval. These methods ought to be considered "safe". This allows user agents to represent other methods, such as POST, PUT and DELETE, in a special way, so that the user is made aware of the fact that a possibly unsafe action is being requested.

ricki-z:
We had to rewrite the WiFiManager lib in a very short time. There were more changes, but i dont upload any single one to github. The actual version uses POST request on all places there this should be done.
Please don't forget, that we work on this project in our (mostly very limited) free time.

corny:
Yes, using POST requests is perfect. But instead of checking the values of submit buttons (where is only one form) it is better to check the request method. This also avoids mistakes like non-matching values. I create my pull requests also in my limited free time ;)

ricki-z:
The actual version shouldn't have a check of the value of submit buttons anymore. Or did is miss one?

corny:
It still checks the value. Your last commit: 19fcb0dea86152bed8c5d937b38d80187fc6f225

PR #58, opened 2017-03-09
Increase serial baud rate to 115200

corny:
It's the default for many ESP8266 as the NodeMCU.


ricki-z:
But its a problem when using longer usb cables. Some NodeMCU have problems with this baudrate. And there is no need for a higher rate.

corny:
Ok, add a comment with this reason to the `Serial.begin(9600);` line.

PR #69, opened 2017-04-25
Set hostname to Luftdaten-$SENSORID

schinken:
It's easier to identify that device in your wifi controller


ricki-z:
I think we will use the new name 'airrohr' in one of the next versions.

ccoenen:
has this ever been implemented? If so, this could be closed. if not: why not merge it?

ricki-z:
we need to update all the docs (or multiple versions) if we change the name. So we may change the name in the future but we need time for preparation if we do this.

YaguraStation:
If you need help with docs, let us know.

z4y4ts:
Just a note from someone who speaks no German.

Luftdaten-XXX would make it much easier to identify the network I need because that's where I've got the Firmware from. That's what I intuitively expected to find.

OTOH, Feinstasomething makes no sense on the first encounter so I have to go back to the documentation to know what to look for.

PR #68, opened 2017-04-21
Fixes #63: Use platformio for enhanced command line development

hakan42:
With this setup, you can 
* clone the repository
* go to the directory esp8266-arduino/ppd42ns-wificonfig-ppd-sds-dht
* build the software with "pio run"
* build and upload the software with "pio -t upload"

If the PR is accepted, I would write another enhancement for the documentation as well.

I had to add a gitmodule reference to TinyGPSPlus which can be removed whenever https://github.com/platformio/platformio-libmirror/issues/99 is resolved.

Issue #64, opened 2017-04-11
Multi-language web interface question


darcoto:
Hi,
Do you have any plans for multi-language web interface of ESP?

I know that I can fork repo, but it is not very useful to  keep it in sync with master only for translation.
I hope you will choose some technology for multi-language strings in sources.
If you are agree that it must be done and give some ideas, I can take some work to implement it.

Best regards

ricki-z:
We can't save too many language versions in the firmware. The FlashROM and RAM ist very limited. We don't have any idea by now how to solve this.

darcoto:
I made some tests to find more effective way to solve. I try some different approach (with ROM and RAM storage) and my conclusion is that it is not very effective to store more then one language in FlashROM without unnecessarily space lost. 
In real usage people never use more than one language in there devices.
So I can offer solution that will build N-firmware files (where N=number of translated languages) each one contains core functions + one language. 
Every language will be one extra .h file and compiler will make N-.bin files with suffix like xxx_de.bin, xxx_en.bin
Auto update must send request with parameter for current language and download relevant file.

I can make changes and suggest pull request, but first of all I would like to know your opinion - whether this approach acceptable for you ?

Best regards

ricki-z:
Can you give me the command line to compile the firmware with platform.io?

hakan42:
pio run

Or, if you want to upload the generated binary to an esp

pio run -t upload

hakan42:
This reminds me that I wanted to write documentation for using platform.io

ricki-z:
I'm getting errors on pio run: .platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld: .pioenvs/nodemcuv2_bg/firmware.elf section `.text' will not fit in region `iram1_0_seg'
There seems to be a problem with the actual version.

darcoto:
Hmm,
That's strange.
Can you give me some information about build environment ? 
Currently I can not reproduce it.
I successfully build 3 firmwares on Mac OS and on Ubuntu with latest version of platformio. Ububtu installation is fresh and clear.   

You can compile single firmware with command
`pio run -e nodemcuv2_de`
`pio run -e nodemcuv2_en`
`pio run -e nodemcuv2_bg`

Try to find is the problem persist in all three version or only in bulgarian.
[...]

PR #75, opened 2017-05-08
Create multi language interface.

darcoto:
According the feature described in #64 I suggest this decision: 

Excluded all texts from main file to external files (named as intl_*.h) and replacing with constants. One file for every language. Currently available three languages - German, English and Bulgarian. 

On build process:

- If build is made with platformio (according this feature https://github.com/opendata-stuttgart/sensors-software/pull/68), the result is  **three firmware files** (firmware_de, firmware_en, firmware_bg.bin)  
- if build is made Arduino, the result is one firmware with German language.

_This approach is most effective for memory optimization against all language in the one firmware._ 

To switch the language user must change the firmware. This can be done very easy via auto update option. Just change current language from Configuration Page and restart the device. On restart device check for new version (if enable) and can download firmware with selected language (I will make another pull request with necessary changes in api.madavi.de/update/firmware.php)

New language can be add very easy. If the PR is accepted, I would write necessary documentation.

Addition changes:
Small UI changes for html optimization and better user expirience 

Please, make a review to this PR and comment. I am ready to change parts that are not suitable to your roadmap. I believe it will help to promote the initiative


darcoto:
To download selected language/firmware on change, this PR must be approved too (https://github.com/opendata-stuttgart/madavi-api/pull/3)

ricki-z:
Will start with that. Firmware version could be changed to NRZ-2017-XXX-DE for german and equivalent for other languages. If LANG is set it could be added to firmware-version on update request.
In the update script the (new) version string could be split in NRZ + year + version + lang. With that we could add a switch in the wifi config for the available languages.

darcoto:
OK.
Just to clarify:
1. "Firmware version could be changed to NRZ-2017-XXX-DE"
Which places:
 - in parameter in update request 
 - on screen, 
 - in config file, 
 - in JSON request sent to API

ricki-z:
I've seen, that you changed the request to the update server. Your solution is much more easier to implement as my first thoughts. So the renaming of the firmware version is obsolete.

ricki-z:
Can't get a working version with platformio. please see #77

PR #82, opened 2017-05-24
Add dutch and added LiquidCrystal_I2C to lib

Cinezaster:


Cinezaster:
I added "nl" as language and fixed an error I got when I tried to build. It was missing a library.

ricki-z:
Thanks for adding the new translation.

PR #93, opened 2017-06-09
Get available network via additional request

darcoto:
This is just an idea for speedup and more stable loading of config page.
## The problem
I noticed that in some situation it is very hard to open config page. I found that the problem persist when available network are mode then 40-45. In that situation, config page make very big output and output process fail. Make some test and confirm that the problem disappears when limit network to 30 or move to in a "quieter" place.

## The Solutions
So, I suggest **two solutions**:

**First:** (in this pull request) contains move logic for getting available WiFi network to independent reqest witch is called from config page via ajax. In this case, config page contains only static form and load very quickly. WiFi list comes after a second. 
If you think the decision is appropriate, I can optimize it more. Addition - we can add extra button that will call list of available Wifi network on click. This can be available in both mode - before first configuration and on normal work. 

**Second solution:** If you think the change is not appropriate, than I suggest just to limit list of available network to (for example) **30 network**. The list is ordered by signal strength, so it is normal you network to be in first 30 network. In current city environment it is a common occurrence to have more than 60 available networks in one place.

Best

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -940,48 +940,12 @@ void webserver_config() {
 		page_content += F("<form method='POST' action='/config' style='width: 100%;'><b>");
 		page_content += FPSTR(INTL_WLAN_DATEN);
 		page_content += F("</b><br/>");
+		debug_out(F("output config page 1"),DEBUG_MIN_INFO,1);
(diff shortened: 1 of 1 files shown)


ricki-z:
I will merge this pull request. But I may change it in some points afterwards.

ricki-z:
I've uploaded a new version. List of networks should only be shown in AP mode. Otherwise the connection will get lost on wifi scan and the sensor will stop working.

PR #129, opened 2017-09-03
Refactor common code to send to dusti into a function.

jms72:
Refactor to remove code duplication.


ricki-z:
Refactored in actual version. Slightly different to show the actual sensor in debug info.

PR #157, opened 2017-11-02
Use released version of TinyGPS++

hakan42:
TinyGPS++ is released on platformio now, use that instead of the specific git hash


ricki-z:
We don't use platform.io for releases. It's not working. Saved config can't be used on update with firmware compiled with platform.io. Compiled binary is larger than the one compiled with Arduino IDE.

PR #181, opened 2018-02-05
Better platformio support + Code formatting for bme280

goggel:


ricki-z:
Hint: we don't really support platformio. Firmware compiled with platformio may destroy a saved configuration.

PR #193, opened 2018-03-22
Support Wemos D1 board and its shields

dok-net:
Besides the compile time guards to support compilation for the Wemos platform, I have changed the structure of the loop() function into a more async, continuations-like style. This makes the system behave more dynamically. I think I have also fixed a few bugs :-)

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1660,7 +1660,7 @@ void beginWifiConfig() {
 	IPAddress netMsk(255, 255, 255, 0);
 
 	debug_out(F("Starting WiFiManager"), DEBUG_MIN_INFO, 1);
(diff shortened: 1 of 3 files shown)


ricki-z:
Please use the Beta branch for pull requests. And the pull request won't be accepted if the styling is changed to your preference. This will give a diff with over 2600 changed rows. There is an astyle file with the used styling.

ricki-z:
Do you read comments????

dok-net:
Calm down. I am just in the final stages of reformatting and trying to merge the beta branch.

ricki-z:
And we won't include most of your changes:
local OTA update (needs flash memory we need for other functions, compiled size has to be smaller than 510 kByte for OTA updates)
sensor handling into library (hint: there is a library for the SDS011 made from this code, installable via Arduino IDE)
yield cleanup (yield is needed to give the NodeMCU some time to handle wifi functions)

PR #198, opened 2018-03-29
Fix for PlatformIO's library selection causing build to fail.

Trickx:
Force use of library ESP8266_SSD1306@3.2.7 in platformio.ini.
Otherwise the builds fail.

### DIFF
--- a/airrohr-firmware/platformio.ini
+++ b/airrohr-firmware/platformio.ini
@@ -29,7 +29,7 @@ lib_deps_external =
   ESP8266HTTPClient
   ESP8266WebServer
   ESP8266WiFi
-  ESP8266_SSD1306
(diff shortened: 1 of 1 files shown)


Adorfer:
Will this make a new stable ("master") possible soon? I would highly appreciate that.

ricki-z:
This won't make a new stable version possible. We use the Arduino IDE for compilation of the firmware binaries and not platform io.
By now we can't get a stable version with the KRACK patches. There seems to be a ploblem with the lwIP lib and our source code.

Adorfer:
so please then make at least a new build "without the krack patches", especially helpful for those who operate the units today in an open network (like freifunk). 
And the improvement on the json/external providers will be hopefully significant compared to Build99.

PR #202, opened 2018-04-12
Add Polish translation of airrohr firmware

monopolka:
Polish translation of the airrohr firmware, new file intl_pl.h and airrohr-firmware.ino modifications


monopolka:
@ricki-z so if you don't mind I'll translate it here: "Zapisz i zrestartuj"

Issue #180, opened 2018-01-08
Optional configuration for tx power


tho-wa:
In case if a external antenna with more gain is used it would be nice to have an ioption to reduce tx power less 20dbi.

ricki-z:
@tho-wa We have included this feature in the beta version. But it seems like this will interfere with the DHT22. With setting the TX power level the DHT22 won't be read at any intervall.

PR #211, opened 2018-05-03
Ability to set wlan TX power level (in dBm)

meneerhenk:
#180


ricki-z:
@meneerhenk I may have to remove this. It seems like this will interfere with the DHT22. With setting the TX power level the DHT22 won't be read at every cycle.

meneerhenk:
I aquired a DHT22 and I will check where this comes from.

PR #215, opened 2018-05-28
GPS Output in HTML and OLED display

saschaludwig:
Added output of GPS values for HTML output on /values and on OLED display.
Optimized OLED display layout.
Also added needed strings for all available translations.


ricki-z:
The development is done in the beta branch. And there the GPS output is added

saschaludwig:
Okay, that wasn't obvious before :) Thank you for the information.

Issue #221, opened 2018-06-09
airrohr-firmware: library versions in Readme.md and platformio.ini


bertrik:
I noticed some differences between the library versions mentioned in the airrohr-firmware/Readme.md file and airrohr-firmware/platformio.ini, compared to the platform.ini file I see the following in Readme.md:
* a newer version of ArduinoJson (5.13.1 vs 5.8.4)
* a newer version of Adafruit BME280 library (1.0.7 vs 1.0.6)
* a newer version of DallasTemperature (3.8.0 vs 3.7.7)
* an *older* version of DHT-sensor-library (1.1.1 vs 1.3.0)
* a newer version of ESP8266_SSD1306 (4.0.0 vs 3.2.7)
* actual version of OneWire (2.3.4 vs unspecified)
* actual version of LiquidCrystal_I2C (1.1.2 vs unspecified)
* actual version of SoftwareSerial (1.0.0 vs unspecified)

I propose I write a patch to fix this and see if it compiles:
a) bring platformio.ini in line with Readme.md, using the newer versions mentioned in Readme.md
b) use explicit version in platformio.ini instead of unspecified ones
b) reorder the list of libraries to put it back in alphabetical order again
c) maybe a developer can look at the DHT-sensor-library that is actually newer in the platformio configuration.

Would a patch like that have a chance to be accepted?

I've become a big fan of platformio myself, for the reason of being able to specify dependencies like these explicitly.

bertrik:
Thanks for merging

PR #222, opened 2018-06-09
Update library versions in platformio.ini

bertrik:
This patch updates library versions in platformio.ini to match the versions mentioned in Readme.md, as discussed in https://github.com/opendata-stuttgart/sensors-software/issues/221

For the Adafruit DHT sensor library, this patch takes special care to use the 1.1.1 version from github instead of the latest which is apparently known to have issues.

For readability, I did not re-order the list of libraries alphabetically yet.

The sensor firmware compiles with this patch, I have not tested it on actual hardware.


ricki-z:
Could you please use the beta branch for pull requests (see README.md)?

bertrik:
Ok, I'll create a new pull request from the beta branch then close this one.

PR #226, opened 2018-06-14
Changing lcd display function

sjo00:


ricki-z:
The esp8266 doesn't have a real time clock. And NTP ist very unstable. So there is no time you could display.

ricki-z:
And again: usage NTP will result in a very unstable firmware !!!!!

sjo00:
Sorry. Thank you for your great work with the airrohr firmware

PR #230, opened 2018-07-02
Improved translation

LuchtwachtersDelft:
Clarified the labels for the optional displays in the English and Dutch translations.
Replaced untranslated strings for "Eigene API:" and "wlanssid" with i18n ones.


ricki-z:
@LuchtwachtersDelft Hello, could you please take a look at the actual version. We need a Dutch translation for "Save and restart".

LuchtwachtersDelft:
How many characters do I get?

Opslaan en herstarten

ricki-z:
Thank you, this is okay. There is no limitation beside the flash size. Texts should be as long as needed and as short as possible.
We have thought about restarting the sensor after saving the config in every case. There are only few cases where we don't need to restart after changes. This that we will save some space.

LuchtwachtersDelft:
Makes me wonder why we need to say "restart". 

Is the "restarting" relevant information to the user who is forced to restart after saving anyway? 

Is mentioning "restart" to warn the user will be a few minutes of downtime because of the warming up phase at startup? If we are restarting when the sensors are already warmed up, why are we warming up again?

If we can streamline the restarting procedure in such a way that the user barely notices it, we might as well remove it from the button.

LuchtwachtersDelft:
I assumed the startup phase was intended for warming up the sensors, but are you saying the actual reason is that the last duration of the last run is unknown? Can it be written to flash when pressing save just before restarting? If the startup time can be reduced that would be a huge improvement for the first impression of new users and for testing (although one can decrease the interval).

I'm curious which reasons other users have for wanting to know that the system will restart. Then we can address those reasons, and perhaps remove some friction here and there.

ricki-z:
I have stopped wondering about the thoughts of some people ...
But telling the user that the sensor is restarting will tell him that the sensor might not be accessible via browser for some time. And this we can't avoid. With older firmware (not showing that the sensor is restarting) we had many users complaining that the page wasn't loading after configuring the sensor. It's not the time until the first measurement that is the problem, this is shown on the page with the actual values.
I'm not sure how often we can really write to the flash memory. So we try to limit this to the absolutely necessary.

PR #250, opened 2018-09-10
Remove unused variables

bertrik:
The compiler shows several compiler errors while building. This patch fixes the warnings about unused variables in the .ino file, by simply removing the variables. The actual executable doesn't appear to get any smaller, but at least it fixes the compiler warnings.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -651,7 +651,6 @@ String SDS_version_date() {
 	String device_id = "";
 	int checksum_is;
 	int checksum_ok = 0;
-	int position = 0;
 
 	debug_out(F("Start reading SDS011 version date"), DEBUG_MED_INFO, 1);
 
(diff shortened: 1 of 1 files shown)


ricki-z:
@bertrik A comment to your commit:
A warning is not an error! It's just a warning, a hint. I.e. an unused variable is listed as warning but removed by compiler optimizations. That's why the executable won't be smaller.
In some cases variables are added but the function that will use this variable is written later. Other warnings may include deprecated functions which may be removed in a future version of i.e. a library. But you can use these functions with the actual version. The warning should only tell you that you may need to change the code when using future versions of this library.

bertrik:
You're right, they are indeed warnings, not errors. I should not have called them errors.

In general I think it's wise to try to keep the number of warnings in code to 0, so you don't need to memorize which ones (or how many) are "harmless" and which ones are actual issues. Sometimes that does mean you need to put in some effort to fix and "issue" that isn't actually an issue yet. Keeping the number at 0 makes it so much easier to spot new ones.

There are some warnings typical to Arduino libraries, which are actually bugs in my eyes to be honest. For example, many Arduino libraries contain functions that are supposed to return a value according to their header file, but the implementation in the .cpp simply returns without a value. I think this results in an undefined value towards the caller. If it were up to me, I would classify these as an error instead of warning :)

PR #261, opened 2018-09-28
Proposal: Read multiple DS18B20 sensors from the OneWire bus

sensor-freak:
I have made minor modifications for the DS18B20 sensor reading to support multiple sensors on the bus.

The change requires the introduction of an additional yield() call during sensor reading, since otherwise the watchdog will force a reset when about 4-5 sensors (or more) are present. I've tested the change with up to 9 sensors now.

In order to distinguish the sensors I have added the twe LSB of the sensors serial number to the display names.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1660,8 +1660,17 @@ void webserver_values() {
 			page_content += table_row_from_value(FPSTR(SENSORS_BME280), FPSTR(INTL_LUFTDRUCK),  check_display_value(last_value_BME280_P / 100.0, (-1 / 100.0), 2, 0), unit_P);
 		}
 		if (ds18b20_read) {
+			const int iNumSensors = ds18b20.getDS18Count();
 			page_content += empty_row;
-			page_content += table_row_from_value(FPSTR(SENSORS_DS18B20), FPSTR(INTL_TEMPERATUR), check_display_value(last_value_DS18B20_T, -128, 1, 0), unit_T);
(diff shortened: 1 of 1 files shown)


ricki-z:
I there any practical need for more than one sensor for the mayority of users? We need to save space for other sensors or extensions. So changes should have a value for our users.
All sensors should be placed outside and our API should only receive outdoor temperature/humidity values. So one DS18B20 should be enough.

ropeters68:
Not for me I use the BME280 to measure outside conditions and the DT22 for the (heated) sensor conditions (more concerned with humidity than temperature).
----Origineel Bericht----
Van : notifications@github.com
Datum : 28/09/2018 09:34
Aan : sensors-software@noreply.github.com
Cc : subscribed@noreply.github.com
Onderwerp : Re: [opendata-stuttgart/sensors-software] Proposal: Read multiple DS18B20 sensors from the OneWire bus (#261)
I there any practical need for more than one sensor for the mayority of users? We need to save space for other sensors or extensions. So changes should have a value for our users.
 All sensors should be placed outside and our API should only receive outdoor temperature/humidity values. So one DS18B20 should be enough.

sensor-freak:
The average user might indeed not want to do this.

I had a different application in mind: measuring the temperatures in my heating system. For this I have places multiple sensors at different places (heat pump flow/return, heating flow/return, room temperature, ...), and of cuorse I do not send this data to the public, but only to a local InfluxDB instance. But you're right, this seems not to be the intention of the project.

So feel free to ignore the pull request.

PR #264, opened 2018-10-10
Remove an unused variable

fmeerkoetter:


fmeerkoetter:
> This variable will be used. It should be the size of one row the array where the wifiInfos will be saved. But there seems to be something missing, so that I have set the row size to a fixed value.

While looking through the firmware I have come (again) across this variable.

I have to admin that I have not understood how you plan to use it (or why you need it)?

ricki-z:
Now we can remove the variable. In older versions of the Arduino IDE (or the esp8266 lib) we needed to create a buffer for this. Stack and heap management were very unstable if we didn't free the memory explicitly. But the actual version seems to have a better memory management.
If you search for 'newer' C++ features and if they are supported by the Arduino IDE you will get many hits with Google search.
Remember the template<> definition where we need to define more than needed to get it compiled with the Arduino IDE.

PR #265, opened 2018-10-10
Remove dead variable

fmeerkoetter:


ricki-z:
As we will activate the validation of certificates for HTTPS connections, we will have a local time.
PLEASE: don't delete "unused" variables in the beta branch. This will be cleaned up for the release version. But until now this may be needed.

PR #266, opened 2018-10-10
Extract duplicated code into a function

fmeerkoetter:


ricki-z:
@fmeerkoetter If you find other parts where we can save some space we would appreciate that ;-) After switching to esp8266 lib 2.4.2 the binary size became considerably larger. And for OTA updates we need to stay below 512 kByte ...

PR #268, opened 2018-10-11
Do not repeat a ternary over and over. Evaluate it once.

fmeerkoetter:
This isolates the "logic" in one place.
Saves 48 bytes.

Second commit saves 128 bytes.
Also note that the control flow for HTTPs is slightly different. When we fail to load to root CA we directly bail out. No need to validate the chain (AFAICS).

DO NOT merge without proper testing.

**not tested on the device (ordered new modules though...)**


ricki-z:
Good to have someone look at the code. After a while it becomes difficult to see some obvious things... ;-)

ricki-z:
seems to work.

PR #272, opened 2018-10-14
Update platformio configuration: add common build flags

bertrik:
This change updates the common build flags for the platformio build platform to match the description for settings in the Arduino IDE, specifically:
    * select low memory lwip2 library
    * vtables in flash
    * no debug output

### DIFF
--- a/airrohr-firmware/platformio.ini
+++ b/airrohr-firmware/platformio.ini
@@ -13,6 +13,9 @@ src_dir = .
 
 [common]
 build_flags =
+  -DPIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY
+  -DVTABLES_IN_FLASH
+  -DNDEBUG
 ; Always depend on specific library versions (wherever possible)
 ; Keep Library names in alphabetical order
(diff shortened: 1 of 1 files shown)


ricki-z:
@bertrik : To save some space in flash the printf compiler switches (in platform.txt) are disabled. We don't need the float support in printf. Is it possible to include this?

bertrik:
I see a reference into the option to disable floats in this thread: https://community.platformio.org/t/declaring-string-variables-and-functions/1835/9
It suggests to add the following to the platformio build_flags: -Wl,-u,_printf_float,-u,_scanf_float
Option -Wl indicates that the rest of the arguments need to be passed to the linker.

I tried it but I don't see a difference in the executable size unfortunately.
Maybe it works a little different than I think/hope, or maybe it works implicitly.

ricki-z:
I would like to use platformio, but I can't get it run on MacOS. It seems not to find the esp8266 lib ...

fmeerkoetter:
@bertrik Might I ask you a question? I am also using platformio (at least for editing). I have the problem, that when I try to compile it will create one firmware image for each language variant of the firmware (which takes quite a bit of time). Can you tell me how to change this? To only built a single image?

PR #274, opened 2018-10-15
Wifi fixes

fmeerkoetter:
**Not tested on the device.**

### DIFF
--- a/airrohr-firmware/.gitignore
+++ b/airrohr-firmware/.gitignore
@@ -9,3 +9,6 @@ builds/*
 !builds/readme.md
 .clang_complete
 .gcc-flags.json
+.vscode/.browse.c_cpp.db*
+.vscode/c_cpp_properties.json
+.vscode/launch.json
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1539,9 +1539,6 @@ void webserver_wifi() {
 		page_content += FPSTR(INTL_KEINE_NETZWERKE);
 		page_content += F("<br/>");
 	} else {
-		page_content += FPSTR(INTL_NETZWERKE_GEFUNDEN);
-		page_content += String(count_wifiInfo);
-		page_content += F("<br/>");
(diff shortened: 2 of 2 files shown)


fmeerkoetter:
@ricki-z I dropped the struct_ renaming.

btw. I now have three additional nodemcus for testing. I also ordered a bunch of sensors (which are still to be delivered). Flashing works fine now.

While playing with the device/firmware I noticed something I would consider a bug.

To configuration website tells me ""wifi networks found: 6". The list it shows only has 3 entries. This is due to "duplicates" being filtered away. IMHO we should tell the user "wifi networks found: 3" and have a matching list. The user cares about networks (everything with the same BSSID is considered the same network), not about individual APs.

fmeerkoetter:
Question: is my assumption correct that the firmware doesn't support
connections to hidden networks?

The code in this function is quite complicated (using this helper array
indices). Do you know if this is really necessary?


Rajko Zschiegner <notifications@github.com> schrieb am Di., 16. Okt. 2018,
10:26:

fmeerkoetter:
Hmm, that was more or less what I had understood. I am still wondering if
moving the structs around is overly expensive. One might consider to do an
insertion sort or something. I'll do some measurements.

About the hidden networks: I will substract them also.

We could add a hint/text if there is at least one hidden network to make
this explicit.

Rajko Zschiegner <notifications@github.com> schrieb am Di., 16. Okt. 2018,
12:24:

fmeerkoetter:
@ricki-z There is still something wrong. In the environment I am currently in the firmware reports "2 networks found" but the list contains four entries... I need to look into that at the weekend...

bertrik:
Perhaps just wondering about a thing I don't really know much about yet, but why would we be concerned about the actual number of networks? We can just show the filtered list, right? The purpose of the firmware is not to be some kind of ultimate wifi network counting device, it's main function is to measure particulate matter.

ricki-z:
Fast solution: not showing the network count
We don't really need this. We should only show a comment if no networks were found.

PR #276, opened 2018-10-16
A branch for platformio tweaks

fmeerkoetter:
@bertrik  ^ jfyi

### DIFF
--- a/airrohr-firmware/platformio_script.py
+++ b/airrohr-firmware/platformio_script.py
@@ -1,11 +1,11 @@
-
 Import("env")
 import os
 import shutil
 from base64 import b64decode
 
 def after_build(source, target, env):
+    if not os.path.exists("./builds"):
+        os.mkdir("./builds")
     shutil.copy(target[0].path, "./builds/latest_"+b64decode(ARGUMENTS.get("LANG"))+".bin")
-
 
 env.AddPostAction("$BUILD_DIR/firmware.bin", after_build)


fmeerkoetter:
```
Compiling .pioenvs/nodemcuv2_en/src/airrohr-firmware.ino.cpp.o
Retrieving maximum program size .pioenvs/nodemcuv2_en/firmware.elf
Building .pioenvs/nodemcuv2_en/firmware.bin
Checking size .pioenvs/nodemcuv2_en/firmware.elf
after_build([".pioenvs/nodemcuv2_en/firmware.bin"], [".pioenvs/nodemcuv2_en/firmware.elf"])
*** [.pioenvs/nodemcuv2_en/firmware.bin] ./builds/latest_en.bin: No such file or directory
Memory Usage -> http://bit.ly/pio-memory-usage
DATA:    [=====     ]  48.2% (used 39460 bytes from 81920 bytes)
PROGRAM: [=====     ]  49.6% (used 518188 bytes from 1044464 bytes)
=========================================================================== [ERROR] Took 3.86 seconds ===========================================================================
```

@bertrik do you know how to resolve this issue?

fmeerkoetter:
Given the last two commits on this branch I have a working platformio setup (which makes me very happy). Good bye Arduino IDE.   :-)

ricki-z:
other option to compile only one version is the CLI: platformio run -e nodemcuv2
In the actual beta branch I have changed the platformio_script.py. So the float build flags are disabled and the size is only 256 Bytes larger than the version generated by the Arduino IDE.

fmeerkoetter:
Ok. Will drop the default env patch. I am not sure if I understood your
answer regarding the python script. Does it already have the mkdir?

Rajko Zschiegner <notifications@github.com> schrieb am Di., 16. Okt. 2018,
22:10:

fmeerkoetter:
the mkdir has been added somewhere else

fmeerkoetter:
> other option to compile only one version is the CLI: platformio run -e nodemcuv2
> In the actual beta branch I have changed the platformio_script.py. So the float build flags are disabled and the size is only 256 Bytes larger than the version generated by the Arduino IDE.

@ricki-z do you have an idea why it is 256 larger? Platformio ships its own toolchain right?

fmeerkoetter:
> other option to compile only one version is the CLI: platformio run -e nodemcuv2

@ricki-z  this is true but with this approach i still can't just press the "upload" button without waiting for ~10 firmwares to be built. For me this only works when setting "env_default". Am I missing something?

ricki-z:
For upload the CLI command is: platformio run -e nodemcuv2 -t upload
Otherwise we need to change the platformio.ini file everytime we want to build all languages. Or need to build every version separately.
And yes, the toolchain is slightly different for platformio.

PR #278, opened 2018-10-17
Replace the "*" with a unicode padlock

fmeerkoetter:
I didn't knew what to make of the "*". I only understood it after reading the
source.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1577,7 +1577,7 @@ void webserver_wifi() {
 				continue;
 			}
 			// Print SSID and RSSI for each network found
-			page_content += wlan_ssid_to_table_row(wifiInfo[indices[i]].ssid, ((wifiInfo[indices[i]].encryptionType == ENC_TYPE_NONE) ? " " : "*"), wifiInfo[indices[i]].RSSI);
+			page_content += wlan_ssid_to_table_row(wifiInfo[indices[i]].ssid, ((wifiInfo[indices[i]].encryptionType == ENC_TYPE_NONE) ? " " : u8"🔒"), wifiInfo[indices[i]].RSSI);
 		}
 		page_content += F("</table><br/><br/>");
 	}


ricki-z:
Should we really use this? Ascii chars up to decimal 127 should be okay on all devices and with most charsets. UTF-8 may cause some problems (freaks with forced charset or encodings for all websites ;-) )

fmeerkoetter:
> Should we really use this? Ascii chars up to decimal 127 should be okay on all devices and with most charsets. UTF-8 may cause some problems (freaks with forced charset or encodings for all websites ;-) )

The websites generated by the firmware are UTF-8 encoded (even though we don't announce it in the <head> section). Browsers assume UTF-8 as default.

For example: 		const String unit_PM = "µg/m³";

Also we set the mime type to be

const char TXT_CONTENT_TYPE_TEXT_HTML[] PROGMEM = "text/html; charset=utf-8";

Also think about the translations. None of these are really. ASCII (even the German one), None of these are encoded in some national charset.

```
 fmee   padlock  …  privat  sensors-software  airrohr-firmware  file *.h
DHT.h:           C++ source, ASCII text
ext_def.h:       UTF-8 Unicode text
html-content.h:  HTML document, ASCII text
intl_bg.h:       UTF-8 Unicode text, with very long lines
intl_cz.h:       UTF-8 Unicode text, with very long lines
intl_de.h:       UTF-8 Unicode text, with very long lines
intl_en.h:       UTF-8 Unicode text, with very long lines
intl_es.h:       UTF-8 Unicode text, with very long lines
intl_fr.h:       UTF-8 Unicode text, with very long lines
intl_it.h:       UTF-8 Unicode text, with very long lines
intl_lu.h:       UTF-8 Unicode text, with very long lines
intl_nl.h:       ASCII text, with very long lines
intl_pl.h:       UTF-8 Unicode text, with very long lines
intl_pt.h:       UTF-8 Unicode text, with very long lines
intl_ru.h:       UTF-8 Unicode text, with very long lines
intl_se.h:       UTF-8 Unicode text, with very long lines
```
The web in 2018 is unicode.


From my perspective the more interesting question is how widespread fonts are with a glyph for a padlock.

I've tested on Android 6 device (which means a two year old firmware) and an Ubuntu 16.04 laptop. Both were fine.

PR #282, opened 2018-10-23
Fix "Read from" text in polish translation

Informatic:
Just add missing template variable in message.


ricki-z:
@Informatic I have added a new text for "save and restart". Could you please translate this.

PR #286, opened 2018-10-26
Better cmd handling

fmeerkoetter:
@ricki-z I hope this is a direction you can agree on. Unfortunately I don't have (spare) sensors of this type available right now. If you consider merging these please **make sure to test. I wasn't able to do so!**


ricki-z:
Do you mean that we should optimize the PM sensor functions that far? The HPM as an example was added only some months before. And not all sensors will have all commands. So we may have a sensor without a continuous mode command or a version date but with something else. At the moment we can copy and paste the code and only need to change some command codes and 'case' lines.

fmeerkoetter:
If such a sensor needs to be supported there is no reason why we can't do
an implementation in a different style (for example switch/case). I don't
see this as a problem.  :-)

My question was a bit about the general direction I am taking (more
modularity, less global state, more typesafe C++, less C style code, small
testable functions, ...). All of that while keeping or even decreasing the
size of the firmware.



Am Fr., 26. Okt. 2018, 22:30 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

ricki-z:
It's 'grown' code so everything what makes this more readable is okay for me.
And also every decrease in the size of the firmware is much apprechiated.
But we have to care about future extensions of the firmware. And this will be mainly new sensors and new languages. Think about you, me and a few other people leaving the project. Someone completely new to this code should be able to add sensors or languages in a forseeable time.
And do we really save space by defining an array of char arrays instead of defining 4 or 5 single char arrays with descriptive names? For the array of arrays we should need more space (pointer to array of pointers).

fmeerkoetter:
@ricki-z I've dropped the commit that remove the switch/case from the patch series. I feel it is the worse solution, but its not a "make or break" for me.  :-)

Still not tested on the device (as I don't have these, or have spares).

fmeerkoetter:
@ricki-z about the general direction. we are on the same page. When I saw the firmware the first time it felt "grown" and I am doing some gardening.

I will work on making it easier to add new sensors (I already have a branch where the are sensor classes and objects). The branch is not complete and ready for prime time.

I am also thinking about unit tests and a travis CI setup.
Another area I started to work on is better logging. The current state is too verbose for my taste.

ricki-z:
We started with two sensors (PPD42NS and DHT22) and a manual config, compiling the firmware for every single sensor. Then the SDS011 was added, the WifiConfig was added, ... And all this in a live system. So we tried to minimize the changes to avoid problems on already working sensors.
At the moment I'm waiting for some translations. With those we can 'simplify' the config functions. In most cases we need to restart the sensor to initiate the hardware or the wifi connection. So we could move the sensor restart on config change into the wifi config function. This will remove some tests and function calls.
Logging with the esp8266 is not really easy (without blowing up the firmware images). In case of the PM sensors we needed to be verbose. Some of them have only little or badly translated documentation. Libraries for them doesn't really existed at the time of adding them to the firmware. As we have only a handful of people testing beta versions we need a possibility to find problems on their test sensors. And for this we need logging messages telling us what the sensor is doing short before a crash or malfunction.

A Travis CI setup should be possible. PlatformIO is supported. But I have to look for the other parts of the 'sensors-software' repo as we can only add one .travis.yml file per repo.

And as you may have seen I have changed the INTL_ var names to english. 

If it's okay for you I could add you to our Luftdaten.info Slack. We have a firmware-dev-channel there. If you would like to join you can send me you mail address to rajko(at)luftdaten.info

ricki-z:
That's okay for me. Answers may need some time here, because notifications go to an email that I use also for all other mails regarding the project ;-)

Informatic:
Just my two cents: More modularity is always a good thing. Furthermore, I think making sensor/peripheral initialization dynamic would allow us to recover *a lot* of memory. (I really doubt anyone would use all PMS5003, SDS011 and HPM* sensors at the same time, or the same goes with SSD1306 + SH1106 + all three LCD displays...)

ricki-z:
@Informatic We use Arduino because of the broad base of people knowing this. In the 3 years of development we had many people working for a short time on this software and then leaving for some other projects. And this project is also used in schools or at universities where Arduino is also more common.
About modularity: could you explain in short what you mean with dynamic initialization?

PR #287, opened 2018-10-27
Timekeeping

fmeerkoetter:
I might be wrong, but I think the forced restart wasn't happening.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -348,8 +348,10 @@ bool send_failed = false;
 
 unsigned long time_for_wifi_config = 600000;
 
+const unsigned long ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
+const unsigned long PAUSE_BETWEEN_UPDATE_ATTEMPTS_MS = ONE_DAY_IN_MS;        // check for firmware updates once a day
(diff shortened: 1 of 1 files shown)


airrohr-firmware/airrohr-firmware.ino
```diff
@@ -3946,13 +3945,14 @@ void loop() {
...
+		const unsigned long ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
 		//forced restart after 4 weeks
-		if ((act_milli - last_update_attempt) > (28 * pause_between_update_attempts)) {
+		if ((act_milli - time_point_device_start_ms) > (28 * ONE_DAY_IN_MS)) {
 			ESP.restart();
 		}
 
 		// check for updates (default once per day)
-		if ((act_milli - last_update_attempt) > pause_between_update_attempts) {
+		if ((act_milli - last_update_attempt) > ONE_DAY_IN_MS) {
```
ricki-z: This is not necessary one day in the future. If there are more and more sensors we may change this to another timespan. Thats why this variable was called pause_between update_attempts.

fmeerkoetter:
I rewrote the patch.

Note that I haven't tested the "forcedRestart" bit. I still think that it wasn't working before.

ricki-z:
You are right, the actual forced restart doesn't work. We should check for act_milli - uptime > 28 days. So your changes should solve this.

PR #288, opened 2018-10-27
Shrinking3

fmeerkoetter:


airrohr-firmware/airrohr-firmware.ino
```diff
@@ -528,10 +527,8 @@ String Float2String(const double value) {
 String Float2String(const double value, uint8_t digits) {
 	// Convert a float to String with two decimals.
 	char temp[15];
-	String s;
-
 	dtostrf(value, 13, digits, temp);
-	s = String(temp);
+	String s = temp;
```
ricki-z: Don't we need to cast temp? It's a char array assigned to a string now.

ricki-z:
Have you seen the Travis CI check ;-) ?

fmeerkoetter:
> Have you seen the Travis CI check ;-) ?

Yes, very nice! Thank you.

airrohr-firmware/airrohr-firmware.ino
fmeerkoetter: No worries.

String s = temp; works because there is a String constructor that takes a "const char".

```
class String {
    public:
        // constructors
        String(const char *cstr = "");
        ...
}
```
The tricky bit is that "String s = "bla"" looks like an assignment and one might suspect the assignment operator of String being called, but this is not the case

ricki-z: As I said, too many different languages ...

ricki-z:
@fmeerkoetter could you please test your changes with the Arduino IDE? After one of the last commits (that was okay with platformio) I can't compile the source with the Arduino IDE.
The Arduino IDE doesn't support the 'enum class' construct.

ricki-z:
@fmeerkoetter and please check that your changes are mergeable. Even if the last commits were mergeable separately, they weren't mergeable at all.

fmeerkoetter:
> @fmeerkoetter could you please test your changes with the Arduino IDE? After one of the last commits (that was okay with platformio) I can't compile the source with the Arduino IDE.
> The Arduino IDE doesn't support the 'enum class' construct.

I wasn't aware that I also have to test with the Arduino-IDE. Since I had a working platformio setup I never touched it. Sorry for that.

I just did and tested. I had no problems with "enum class". It compiled fine with the Arduino-IDE.
I've re-opened https://github.com/opendata-stuttgart/sensors-software/pull/291 not to annoy you, but so you have an isolated patch you can try.

Can you please post the error message (if you encounter errors)?

ricki-z:
Platformio produces larger firmware images. As long as we don't know the source of those extra bytes, we should compile with the Arduino IDE. And sometimes platformio is changing its behaviour in an unpredictable way. Or things doesn't work like expected. I.e. there is a difference in the commands run and ci which leads to problems with extra_scripts. But we need this script to disable some linker flags.

fmeerkoetter:
> I just did and tested. I had no problems with "enum class". It compiled fine with the Arduino-IDE.
> I've re-opened #291 not to annoy you, but so you have an isolated patch you can try.

@ricki-z ignore me. I reverted your revert - and now I have an error. Looking into it.

ricki-z:
I got an error like 'type PmSensorCmd is not defined'. Now it seems to work (actual beta branch). As you can see this was the only thing I changed back. I could find some information that the order of definitions may cause problems.

bertrik:
@ricki-z If we have elf images for both platforms (Arduino IDE and platformio), I think I can try running them through a script I used on another project. It basically runs them through 'nm'. This should give me more insight on where the differences are and what causes the size differences. I hope to find some time for this the coming week.

ricki-z:
@bertrik may I send you the two files via email? I have found an address from your sensor registration that I could use.

PR #294, opened 2018-10-30
return early of webserver_request_auth() "failed"

fmeerkoetter:
while browser wise the everything seems fine and the
basic auth dialog is brought up, I can see on the serial
log that the firmware keeps executing on the "authenticated path"
even though we just sent a 403 to the browser. This is at
least dangerous. Return early.


fmeerkoetter:
Yes, I did test on the device(this is where I noticed the logoutput that
shouldn't be there). But please take your time, I am in no hurry to get
this merged. I would appreciate testing by you as I don't have that much
experience with the firmware and might have missed something.

Am Mi., 31. Okt. 2018, 01:22 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

PR #296, opened 2018-10-31
Cleanup dht read

fmeerkoetter:
Tested on the device

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -2408,7 +2408,7 @@ static String sensorDHT() {
 			debug_out(String(FPSTR(SENSORS_DHT22)) + FPSTR(DBG_TXT_COULDNT_BE_READ), DEBUG_ERROR, 1);
 		} else {
 			debug_out(FPSTR(DBG_TXT_TEMPERATURE), DEBUG_MIN_INFO, 0);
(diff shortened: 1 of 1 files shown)


ricki-z:
The false flag was for a forced read in earlier versions. In newer versions this is the second flag with the first flag for Celsuis or Fahrenheit.
So this call should be`dht.readTemperature(false,true); `
But this needs to be thouroghly tested as there were many problems with the DHT lib. Normally the DHT is read only once per 2 seconds. Within this 2 seconds the sensor is read only with the forced flag. So if there is no result we would try the forced read.
Until now I would like to leave this pull request open.

fmeerkoetter:
> The false flag was for a forced read in earlier versions. In newer versions this is the second flag with the first flag for Celsuis or Fahrenheit.
> So this call should be`dht.readTemperature(false,true); `
> But this needs to be thouroghly tested as there were many problems with the DHT lib. Normally the DHT is read only once per 2 seconds. Within this 2 seconds the sensor is read only with the forced flag. So if there is no result we would try the forced read.
> Until now I would like to leave this pull request open.

Good point. I've dropped this particular commit from the series so you can consider them on your own.

I will create a pull request with a dht.readTemperature(false,true); just that we don't forget it.

fmeerkoetter:
I think I have removed the problematic commit from this series. Could you
please consider pulling the rest?

Am Mo., 5. Nov. 2018, 00:13 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

PR #303, opened 2018-11-04
Limit the form input to 32 chars

fmeerkoetter:
as the underlying char array is 33 chars (32 + null termination)


fmeerkoetter:
Please ignore. Will push a better fix soon

PR #305, opened 2018-11-05
Cleanup booleans

fmeerkoetter:


ricki-z:
Isn't `false && auto_update` false in every case? This would disable auto updates.

PR #308, opened 2018-11-07
Collect all the persistent configuration variables inside a namespace

fmeerkoetter:
The variables inside the cfg namespace are persistent configuration values.
They have defaults (extdef), they can be changed by the user via the web
interface, the changes are persisted to the flash and read back after reboot.

Wrapping these inside a namespace clearly communicates that a given global
variable is actually a special variable (a cfg variable).

This commit is part of an ongoing effort to give the firmware a bit more structure.

This change does not increase the size of the firmware.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -427,8 +427,6 @@ String last_data_string = "";
 
 String esp_chipid;
 
-String server_name;
-
 long last_page_load = millis();
 
 bool wificonfig_loop = false;
@@ -2036,9 +2034,6 @@ void webserver_not_found() {
  * Webserver setup                                               *
  *****************************************************************/
 void setup_webserver() {
-	server_name = F("Feinstaubsensor-");
-	server_name += esp_chipid;
-
 	server.on("/", webserver_root);
 	server.on("/config", webserver_config);
 	server.on("/wifi", webserver_wifi);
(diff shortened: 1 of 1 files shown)


ricki-z:
@fmeerkoetter
'Remove a global variable' server_name should be made configurable or translateable for international versions in a later release. First tests with that showed some problems. But this was left as a remark.
'Demote first_csv_line from a global into a local static' won't work this was. Setting the variable to true one line in front of the condition makes this always true. So the csv header line would be printed in front of every value line. But it should only be printed once per start of the sensor. So the variable needs to remain global to 'survive' the function call.

fmeerkoetter:
Regarding the demote_csv: the variable is not always true.

Please consider the static storage modifier. It has the effect that the
variable will be initialized with true the first time the code is executed.
If the function is executed a second time (3rd, 4th, ...) it is already
initialized and the true assignment won't happen again.

IMHO the code works

Am Do., 8. Nov. 2018, 23:05 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

fmeerkoetter:
Regarding the server name: I will change the patchsets to make it a global
again, adding your remark why it should be a global as comment.

Am Do., 8. Nov. 2018, 23:05 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

ricki-z:
Regarding the demote_csv: this is C ... ;-) As I mentioned before, too many different languages at the same time.

PR #309, opened 2018-11-09
Adapt to a changed API of the dht sensor

fmeerkoetter:
False now means Celsius...
We need to call readTemperature(false, true) to force a read.

Do _NOT_ merge it. This PR is mostly reminder. This needs more testing before merge.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -2415,7 +2415,7 @@ static String sensorDHT() {
 		if (isnan(t) || isnan(h)) {
 			delay(100);
 			h = dht.readHumidity();
-			t = dht.readTemperature(false);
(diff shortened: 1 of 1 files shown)


bertrik:
Oh, wow, such a horrible interface for the DHT sensor ...

I agree that it would probably be better to pass 2 arguments like in your patch, instead of just passing 1 and make a guess which one is overridden by the default.

The DHT library has been known to give some problems between different versions, perhaps this was the reason.

fmeerkoetter:
I was considering upstreaming a patch changing the bool, bool to two
strongly typed enums. Haven't gotten around

Am So., 25. Nov. 2018, 12:36 hat Bertrik Sikken <notifications@github.com>
geschrieben:

dirkmueller:
this is no longer needed, the forked DHT library we use now no longer has this api.

PR #311, opened 2018-11-10
The size of an SSID is 32 + 1 for null termination

fmeerkoetter:
Replace magic numbers with a named constant.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -162,6 +162,7 @@
 /******************************************************************
  * Constants                                                      *
  ******************************************************************/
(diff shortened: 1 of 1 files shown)


ricki-z:
We had set the SSID length to 33 in a future version. But then there were some APs with with names max. SSID lenght, where the sensor couldn't connect anymore. Please don't ask me why (utf8 BOM?).
So we set the length to 35. That was working for those APs.
There seems to be a very different understanding about standards by some of the hardware manufacturers.

ricki-z:
@fmeerkoetter can you set the length to 35, just to be sure?

dirkmueller:
This has been fixed via some other comment already.

PR #314, opened 2018-11-10
Require username AND password to be set

fmeerkoetter:
@ricki-z this is more a question or request for comment than a PR.

I would assume that both username and password need to be non-empty strings (not one OR the other). Do you know?

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1094,11 +1094,11 @@ void writeConfig() {
  *****************************************************************/
 void create_basic_auth_strings() {
 	basic_auth_custom = "";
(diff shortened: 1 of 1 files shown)


ricki-z:
Its unusual but allowed that one of both is empty. Empty user or password is something like a pre-shared key (you only need to remember one 'word').

ricki-z:
@fmeerkoetter can we close this?

fmeerkoetter:
I wanted to turn this into a commit adding a comment. I think I've found
the right RFC, but not the specific wording allowing for one of
username/password being empty.

Feel free to close it. I don't have the time to follow through on this ATM

Am Mi., 21. Nov. 2018, 01:45 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

PR #316, opened 2018-11-12
Introduce msSince()

fmeerkoetter:
This macro hides an ever occuring (simple) calculation behind
a function like syntax (with a speaking name).

Code like

if (msSince(last_update_attempt) > PAUSE_BETWEEN_UPDATE_ATTEMPTS_MS) {

can be read as

milli seconds since last update attempt is larger as the pause in milliseconds
between update attempts.

I would have prefered a function over a macro but there was a significant
increase in firmware size (which I can't explain at this point). This is something
I might tackle later.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -459,6 +459,8 @@ template<typename T, std::size_t N> constexpr std::size_t capacity_null_terminat
 	return N - 1;
 }
 
+#define msSince(timestamp_before) (act_milli - (timestamp_before))
+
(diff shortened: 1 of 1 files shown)


ricki-z:
function call compared to substraction. A substraction are only 3 instructions (push, push, sub), the function call is the same plus the jump to the function.

fmeerkoetter:
@ricki-z hmmm strange. even when forcing the inliner with __attribute__((__always_inline__)) unsigned long msSince(unsigned long) {...} the result is slightly larger (~80 bytes).

PR #319, opened 2018-11-13
Refactor loop

fmeerkoetter:


ricki-z:
@fmeerkoetter what was the problem with encoding (last commit aecb0fc). Just to avoid it in the future. I can't see any difference in the source view.

fmeerkoetter:
Good question. Platformio showed it as an encoding error. The git diff
showed an Unicode code point. Which platform (host) are you working on?

Am Mi., 14. Nov. 2018, 16:19 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

ricki-z:
I'm working with MacOS most time. But maybe someone copied the links on a Windows machine into the source code.

PR #320, opened 2018-11-14
Refactor/remove copyExtDef()

fmeerkoetter:
copyExtDef() isn't really needed. The cfg data was initialized twice - once
via direct initializers, once via copyExtDef(). The direct initializer
weren't used the defines in ext_def.h.

This commit saves 688 bytes.

@ricki-z initial testing on the device looks good. I hope I haven't overlooked something.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -174,73 +174,82 @@ const unsigned long DURATION_BEFORE_FORCED_RESTART_MS = ONE_DAY_IN_MS * 28;  // 
 
 /******************************************************************
  * The variables inside the cfg namespace are persistent          *
- * configuration values. They have defaults (extdef),             *
- * they can be changed by the user via the web interface, the     *
+ * configuration values. They have defaults which can be          *
+ * configured at compile-time via the ext_def.h file              *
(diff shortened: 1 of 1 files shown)


fmeerkoetter:
Hmmm, I might be missing your point - but what you describe still works. :-)

Check out the last two variables in the cfg namespace. These are
initialized directly via a literal, not via a define taken from ext_def.h

Am Mi., 14. Nov. 2018, 23:25 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

fmeerkoetter:
I don't think my patch is actually removing functionality. It just removed
a step where we copy values (all known at compile time) at runtime (which
doesn't make too much sense). I might be missing something?

Am 15.11.2018 6:08 vorm. schrieb "Frank Meerkötter" <frank@meerkoetter.org>:

Hmmm, I might be missing your point - but what you describe still works. :-)

Check out the last two variables in the cfg namespace. These are
initialized directly via a literal, not via a define taken from ext_def.h

Am Mi., 14. Nov. 2018, 23:25 hat Rajko Zschiegner <notifications@github.com>
geschrieben:

ricki-z:
Yes, my fault. It was too late yesterday.

PR #321, opened 2018-11-17
zDo not set the sample count to 0 at the end of each loop

fmeerkoetter:
We want to count the samples, reset sample_count to 0 at the end of each loop doesn't help.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -4026,7 +4026,6 @@ void loop() {
 		last_data_string = data;
 		lowpulseoccupancyP1 = 0;
 		lowpulseoccupancyP2 = 0;
-		sample_count = 0;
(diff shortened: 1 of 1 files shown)


ricki-z:
This counts how often the loop is executed per measurement interval. Complete time needed for an interval divided by this is the average time per run through the loop.
We need this for the PPD42NS where we need to know how often the pins are read. And these are samples.

ricki-z:
And it isn't set to zero after every loop. It set to zero only if send_now is true (at the end of each measurement interval).

fmeerkoetter:
You are absolutely right. I missed that additional scope. Makes more sense now ;-)

ricki-z:
But maybe we could put the sending lines to a function to get the loop() shorter. In a second step we could move all inside the "if (send_now)" to an own function. So the structure should be more clear.

PR #322, opened 2018-11-17
Extract a bit of code from display_values()

fmeerkoetter:


ricki-z:
Code should be smaller if you only replace the if-then-else with the C-like "expression?true-val:false-val", without a function.

PR #329, opened 2018-11-20
Wait on reading response from server (Fixes #327)

### LINKED ISSUE #327: NRZ-2018-115: submitting data to influxdb stopped with http 499
suddenly the pushing of data to influxdb stopped:

```
$ip - - [19/Nov/2018:19:36:23 +0000] "POST /<censored>/write?db=luftdaten HTTP/1.1" 204 0 "-" "-"
$ip - - [19/Nov/2018:19:39:27 +0000] "POST /<censored>/write?db=luftdaten HTTP/1.1" 204 0 "-" "-"
$ip - - [19/Nov/2018:19:43:19 +0000] "POST /<censored>/write?db=luftdaten HTTP/1.1" 499 0 "-" "-"
```

has there been a change in the code? my sensor updated at the same time to the new build, and the only thing that is special is that I don't use basic auth - I use a random path as authentication, and the influxdb is behind a nginx that just proxies the requests.

dirkmueller:
Give the server some time to get back to us with a reply, otherwise
in case of proxied connections via nginx, the proxy can fail
the request when the client disconnects too quickly. Also we might
not actually see the reply.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -2193,7 +2193,13 @@ void sendData(const String& data, const int pin, const char* host, const int htt
 
 		client->println(data);
 
-		delay(10);
+		// wait for response
+		int retries = 5;
+		while (client->connected() && !client->available()) {
+			delay(10);
+			if (!--retries)
+				break;
+		}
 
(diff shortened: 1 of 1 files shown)


dirkmueller:
Warning: completely untested!

PR #347, opened 2018-12-23
Update platformio configuration to run the ESP8266 at 160 MHz.

bertrik:
I noticed you added a remark that the Arduino binaries are supposed to be built with a CPU frequency setting of 160 MHz. This change makes the same change for platformio.


ricki-z:
Thank you, I was looking for this build .switch

PR #355, opened 2019-01-04
Set the platformio programming upload speed to 921600 bps

bertrik:
This improves the programming time from about 1m4s to about 24s when building and uploading using platformio, saving a lot of time in the development process.

This is done by creating a variable upload_speed in the 'common' environment and referring to this variable in all of the other platformio environments for each language.

### DIFF
--- a/airrohr-firmware/platformio.ini
+++ b/airrohr-firmware/platformio.ini
@@ -44,6 +44,7 @@ lib_deps_external =
   Wire
   LiquidCrystal_I2C@1.1.2
 extra_scripts = platformio_script.py
+upload_speed = 921600
 
 [env:nodemcuv2]
 lang = DE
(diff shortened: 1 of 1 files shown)


ricki-z:
Did you test this with different NodeMCUs and usb cables (different lengths). Even the default upload speed is causing problems with some NodeMCUs that can be solved by decreasing the upload speed. So it's very likely that more NodeMCUs may have problems with an upload speed that high.

bertrik:
I use this setting on all of my esp8266 boards (typically nodemcu and wemos d1 mini). Some USB cables don't work, but then the problems are not just limited to the serial speed but to stability in general and I strongly suspect in those cases it's the resistance of the USB power wires causing the problems. It works for me with the 2 meter "noodle" type of flat USB cable. The higher serial speed only takes place between the USB-serial converter and the esp8266.

So no, I haven't tested this extensively and I understand your concerns, but even with the "standard" 2m long USB cable this works fine for me and saves me a lot of time when experimenting with the code.

PR #357, opened 2019-01-05
support precipitation sensor [WIP]

jmberg:
Precipitation sensors typically just give a pulse when the rocker
flips over due to the weight of the rain water, typically using a
reed contact.

Support connecting such a sensor to D8/+3.3V - we can use D8 even
though it's GPIO15 because it has a built-in pull-down and we just
want to sense the high pulse when the rocker flips.

Allow configuring the amount of rain detected for each pulse.

I've tested this using
https://www.aliexpress.com/item/Free-Shipping-Spare-part-for-weather-station-to-measure-the-rain-volume-for-rain-meter-for/32793155455.html
which has a 0.3mm rocker content.

NB: This commit is woefully incomplete.
 * language updates are missing other than German
 * maybe it should be able to report this to the official APIs?
 * documentation updates are missing
 * sensor name should perhaps not be PULSE
 * software version needs to be updated

As a result, I fully expect you to reject this - but I'd like to see if you're
interested in this at all, and if so what changes I'd need to make.


ricki-z:
First of all: please do your pull requests to the beta branch as mentioned in the main README.md . We are only merging requests there.

PR #358, opened 2019-01-07
add config option to set Wifi host name

bb-Ricardo:
* defines wifi hostname
* defines mdns name

PR #362, opened 2019-01-10
add config option to set Wifi host name

bb-Ricardo:
* defines wifi hostname
* defines mdns name
* sets page title

see #358


ricki-z:
There is an existing option to set the name of the sensor access point in config mode. This name could also be used for mDNS.
And do we really need the name of the sensor in the page title? We try to avoid every single unneeded byte.

YaguraStation:
If I'm not confused, then I think Rajko is talking about "WiFi Sensor in configuration mode" in the section "WiFi Sensor" of the configuration.

PR #369, opened 2019-01-24
Renamed .ino to .cpp to enable code completion and code linting

LuchtwachtersDelft:
Renamed .ino to .cpp to enable code completion and code linting with PlatformIO. See https://docs.platformio.org/en/latest/faq.html#convert-arduino-file-to-c-manually

Reordered functions due to C++ requirements.

Added a link to sensor graph at Madavi.de on the home page. Need to change the other language files.


ricki-z:
This pull request will be closed. With platformio and atom you will have code completion for .ino files. Changing the the file extension would  break compiation with the Arduino IDE.

Issue #383, opened 2019-02-19
Acces Point mode: wifi password transmitted unencrypted? Maybe a security problem


speendo:
A friend and me, we are thinking about building a Feinstaubsensor, therefore I had a short look at the project and the code.

In the first setup, when you have to enter the settings for your home wifi, is this information transmitted encrypted?

On first sight it seems to me that it is not.

In this [line](https://github.com/opendata-stuttgart/sensors-software/blob/7becd7643740820514bb6d64ab9f167685a165a6/airrohr-firmware/airrohr-firmware.ino#L2033) the AccessPoint is established with a password stored in `cfg::fs_pwd` however, as this [line](https://github.com/opendata-stuttgart/sensors-software/blob/7becd7643740820514bb6d64ab9f167685a165a6/airrohr-firmware/ext_def.h#L15) shows, `cfg::fs_pwd` is an empty string.

Therefore, it seems to me that an attacker that "listens" to the communication between me and the ESP8266 while entering the settings for my home wifi could get access to my network like this.

Maybe I missed some other security measures you have taken. If not, an easy solution would be to set `cfg::fs_pwd` to something else than an empty string.

I've also noticed this issue: https://github.com/opendata-stuttgart/sensors-software/issues/128 however, I couldn't see if this particular problem was already solved with the measures you were taking there.

ricki-z:
The password is transmitted unencrypted on send.
What can we do?
- using HTTPS: this would need a certificate, but self-signed certs are throwing a warning (confusing most users)
- using WPA2: we need to set a random password. But how should we tell this password to the user? Not all users are able to read the console output. If you think you will have to change the password regularly, then you can set an own password for the sensor in AP mode

Normally the wifi password should be transmitted only once unencrypted at the moment of saving (and only if changed). In all cases where a 'password' is shown in the input fields only a string with asterisks is transmitted. So someone needs to capture the traffic in exactly this timespan where you configure the sensor the first time.

If you want to avoid the the unencrypted transmission you could compile and flash a modified version of the sensor software where you can predefine your wifi password.

mdoersam:
Just setup two sensor-systems and realized this issues exists right after. I totally agree with speendo. Just "Password" as a password would be a big improvement as it prevents sniffing of the actual wifi-password while it is being transmitted.

This still leaves the issue that someone could be able to reconfigure the sensor, but I don't really see the security-issue with that, as it runs with default configuration at that point

speendo:
[...]

PR #397, opened 2019-04-01
Encrypting default WiFi connection after setup

speendo:
As described in https://github.com/opendata-stuttgart/sensors-software/issues/383 there is a serious security issue when setting up the sensor for the first time.

The credentials of your home WiFi (or whatever WiFi you try to connect your sensor to) are transmitted in plain text so that a potential attacker can sniff them without any effort. Once inside your network the attacker could do a lot of harm to other devices on your network.

To overcome this problem, I introduced a default password "ParticulateMatter255" in `ext_def.h`.

Furthermore, I wanted to avoid that an attacker can easily find out this password by just reading the AP's SSID (when finding the default SSID "Feinstaubsensor-<chipid>" the attacker would only have to search this word on google to find the default password which would enable him/her to decrypt the credentials again...). Therefore I renamed the SSID to the more generic term ESP<chipid> which is used more widely and therefore wouldn't allow an attacker to directly deduct which password to use for decryption.

This is not a perfect solution but for ~90% of usecases this should be safe enough and in any case much safer than the actual solution of not having any encryption on the default access point.

Keep in mind that the manual(s) for setting up the sensor need to be adapted to these changes.

### DIFF
--- a/airrohr-firmware/Readme.md
+++ b/airrohr-firmware/Readme.md
@@ -23,7 +23,7 @@ ppd42ns-wificonfig-ppd-sds-dht.spiffs.bin	-	Binary mit leerem Dateisystem, zum L
 ## WLAN Konfiguration
 siehe auch Wiki-Seite auf Github [Konfiguration der Sensoren](https://github.com/opendata-stuttgart/meta/wiki/Konfiguration-der-Sensoren)
 
-Wenn das vorgegebene WLAN nach 20 Sekunden nicht erreichbar ist, wird ein Access-Point eingerichtet, der über "Feinstaubsensor-\[Sensor-ID\]" erreichbar ist. Nach dem Verbinden zu diesem Accesspoint sollten alle Anfragen auf die Konfigurationsseite umgeleitet werden. Direkte Adresse der Seite ist http://192.168.4.1/ .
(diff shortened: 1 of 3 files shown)


ricki-z:
Could you please push to the beta branch (as mentioned https://github.com/opendata-stuttgart/sensors-software !!!). And no, we won't merge this pull request as we are working on this. That we have to change all the instruction was mentioned in more than one place. And we have to do this in multiple languages, what is causing an additional delay for translations.

speendo:
I looked here https://github.com/opendata-stuttgart/sensors-software/blob/master/contributing.md (I think this page was last edited by you) and didn't find any instructions on how to do pull requests. Never mind, I can redo the pull request to the beta branch as well. Anyway, if you don't consider merging my pull request it doesn't make sense to push it to another branch, does it?

However, no need to be rude. My issue https://github.com/opendata-stuttgart/sensors-software/issues/383 was not discussed further, which I interpreted to be caused by a lack of time to do it. So I decided to try and contribute.

It took quite some time to make everything compile correctly (by the way SoftwareSerial cannot be downloaded inside of Arduino IDE any more - opposed to the instructions here: https://github.com/opendata-stuttgart/sensors-software/tree/master/airrohr-firmware#verwendete-bibliotheken-f%C3%BCr-esp8266).

No hard feelings, but I expected a little more appreciation for people spending their free time to improve your project.

ricki-z:
No hard feelings, but we also expect some appreciation as we do this whole project unsalaried in our leisure time. We organize workshops, develop the firmware and map software, pay and administer all needed servers, add new sensors (beside this we try to develop a self-service portal to add new sensors) and so on. So sorry that we might not have seen this issue or had the time to answer this.

PR #402, opened 2019-04-05
Moving string literals to PROGMEM

ChristianAmmer:
This pull request is for moving string literals to PROGMEM. Additionally duplicated `F` macro string literals are also removed through the `FPSTR` macro. The `wifiInfo` pointer is now owned by a smart pointer and automatically destroyed.

|    | Branch beta | After this PR |
| --- | --- | --- |
| Sketch used program storage space | 579676 | 579396 |
| Global variables use | 35016 | 34984 |

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -254,26 +254,26 @@ namespace cfg {
 	}
 }
 
-#define HOST_MADAVI "api-rrd.madavi.de"
-#define URL_MADAVI "/data.php"
+const char HOST_MADAVI[] PROGMEM = "api-rrd.madavi.de";
+const char URL_MADAVI[] PROGMEM = "/data.php";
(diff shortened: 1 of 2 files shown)


ricki-z:
I have checked some of the replacements of strings with FPSTR. For this I replaced six BR_STR (would be replaced with <br> at compile time) with FPSTR(BR_STR). Result: no difference in image size.
with BR_STR:
image size 568084 Bytes (54%)
global variables 35624 Bytes (43%)

with FPSTR(BR_STR):
image size: 568084 Bytes (54%)
global variables: 35624 Bytes (43%)

FPSTR will save space only for strings larger than 4 bytes. But it may need some more tasks (function call, copy from flash)?

ChristianAmmer:
Hello Rajko
Good to know that. I will remember it for future changes, thank you for trying out.
Also thanks for reviewing and merging this PR.

PR #406, opened 2019-04-16
Fix for aircms data sending function.

DeeKey:
Some data values will not be uploaded otherwise.


ricki-z:
The latest version should contain this change.

PR #413, opened 2019-06-05
Restore SSL support for the custom API

trekawek:
The commit ca7db7531c54a5d3ec3c6bf7fcf649d5c5dc1987 to the beta branch introduced a regression. Before the commit, it was possible to use ssl when connecting to the custom API, by setting the custom API port to 443. After the commit, it's no longer possible - plain http is always used, ignoring the port number.

This PR restores the original behaviour.


trekawek:
//cc: @bb-Ricardo

bb-Ricardo:
Damn, my fault, sorry.

trekawek:
No problem. Let's just fix before merging the beta to master :)

ricki-z:
How likely it it that someone would use port 443 without SSL?
I could add an option for HTTPS and check if the port is 43 OR the option is set.

nikilase:
Is it possible to have an additional checkbox to decide weather SSL is used or not? 
Unfortuneatly we can't use Port 443 but need HTTPS.

ricki-z:
The actual beta has an additional checkbox for SSL for the custom API.

Issue #297, opened 2018-11-02
Add Support for AMS CC811


raspyweather:
Please add support for this cheap air quality sensor

Adorfer:
What does this sensor actually measure? 
From the PDF i deduct, it's 
1) CO2 in the range 400ppm up to 32768ppm
2) "equivalent Total Volatile Organic Compound" (eTVOC) in the range 0ppb up to 32768ppb. 

https://ams.com/documents/20143/36005/CCS811_DS000459_6-00.pdf/c7091525-c7e5-37ac-eedb-b6c6828b0dcf

ropeters68:
That would be a very nice feature!

ricki-z:
This sensor uses some kind of 'auto calibration' using the lowest measured value in the last 24 hours as a baseline. So you can't compare two sensor places because they may have very different 'baselines'.

ricki-z:
This is a problem for the BME680 also.
One possible solution: do all CC811 (or BME680) in one room for 24 hours, save baseline, repeat this every 3 months. But who should do this?

ropeters68:
You can't compare the values even if they have the same baseline because it's an aspecific sensor (you measure something but you don't know what). However you can use it to detect pollution incidents. These type of sensors are widely used in E-Nose projects (like in the Port of Rotterdam and near chemical plants) to detect bad smells and spills. So although the measurments are not directly comparable they do provide valuable data.
----Origineel Bericht----
Van : notifications@github.com
Datum : 09/11/2018 14:52
Aan : sensors-software@noreply.github.com
Cc : roderick.peters@telfort.nl, comment@noreply.github.com
Onderwerp : Re: [opendata-stuttgart/sensors-software] Add Support for AMS CC811 (#297)
This sensor uses some kind of 'auto calibration' using the lowest measured value in the last 24 hours as a baseline. So you can't compare two sensor places because they may have very different 'baselines'.

raspyweather:
TVOC can measure stuff like fumes from paint, chemicals, gasoline etc.. 
You should read this with a grain of salt, but that's a very interesting source about VOCs IMHO:
http://key-vocs.eu/sites/default/files/ENV56%20Final%20Publishable%20Summary_accepted.pdf 

Of course, I'm not a professional in Chemistry and I'm still getting used to measuring my environment and trying to get to know which values could have which background. 

[...]

PR #424, opened 2019-07-11
Adds support for AMS CCS811 CO2 sensor.

mgaggero:
Implements #297 
Adds support for AMS CCS811 CO2 sensor.


ricki-z:
Sorry, we won't add this to our firmware. From the Adafruit shop page https://www.adafruit.com/product/3566:
Please note, this sensor, like all VOC/gas sensors, has variability and to get precise measurements you will want to calibrate it against known sources!
Do you (and all the users of this sensor) have a "known source".

PR #436, opened 2019-08-17
Update adafruit libraries to latest release

dirkmueller:


.travis.yml
```diff
@@ -12,7 +12,7 @@ env:
     - PLATFORMIO_CI_SRC=./
 
 install:
-    - pip install -U platformio
+    - pip install -U platformio<4.0.0
```
bertrik: What's wrong with platformio 4.0.0 and higher?

dirkmueller: See.commit.message. the new version rejects the lang env which causes build to fail in travis.

I forgot to add quotes so it now also fails. Will fix when I am back at the computer

bertrik: Ah ok, I've been seeing warnings about the 'lang' variable for a long time. If this is now an error in platformio 4.0.0, might be best to stick to the older version for now, so we can fix the underlying problem in the meantime.

dirkmueller: As you can see this seems to fix the Travis failure.

dirkmueller: s,fix,workaround,

dirkmueller:
@here any chance that we can merge this?

ricki-z:
Sorry, we had some holiday.
And there were some major changes to the main script that needed to be pushed before.

PR #437, opened 2019-08-22
Update airrohr-firmware.ino

n0c0n:
Odliczanie sekund do pierwszego i od ostatniego pomiaru.
Counting down since last and to first measurement.


ricki-z:
Do we really need a "live" counter for this? This changes need some hundred bytes ot flash we could use for other things like new sensors.
Most users don't use this page regularly.

ricki-z:
So I can close this?

n0c0n:
Yes, ok. If bytes are needed to something more useful, so be it.

PR #442, opened 2019-08-31
Skip erratic readings from SDS011 sensor

dirkmueller:
When the sensor fan is failed, the readouts will be consistent
999 and 1999, which appears to be indicating some kind of error
encoding (unfortunately the datasheet is not mentioning this).
    
Skip those values as they pollute the sensor map with incorrect
data (which appears as huge spikes in pollution in maps)

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1325,13 +1325,6 @@ static String tmpl(const String& patt, const String& value) {
 	return s;
 }
 
-static String tmpl(const String& patt, const String& value1, const String& value2) {
-	String s = patt;
-	s.replace("{v1}", value1);
-	s.replace("{v2}", value2);
-	return s;
-}
-
 static String tmpl(const String& patt, const String& value1, const String& value2, const String& value3) {
 	String s = patt;
 	s.replace("{v1}", value1);
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -4302,11 +4302,9 @@ static unsigned long sendDataToOptionalApis(const String &data) {
 
 		String aircms_data = "L=" + login + "&t=" + String(ts, DEC) + "&airrohr=" + data;
 		String token_hash = sha1Hex(token);
-		String hash = hmac1(String(token_hash), aircms_data + token);
-		char char_full_url[100];
-		sprintf(char_full_url, "%s%s", URL_AIRCMS, hash.c_str());
+    String full_url = URL_AIRCMS + hmac1(String(token_hash), aircms_data + token);
 
-		sendData(aircms_data, 0, HOST_AIRCMS, PORT_AIRCMS, char_full_url, true, false, "", FPSTR(TXT_CONTENT_TYPE_TEXT_PLAIN));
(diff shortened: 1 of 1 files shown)


ricki-z:
This should be filtered by the map. We need these values so that we can give a hint to the users (work in progress).

ricki-z:
For single erratic values there is some filtering as we throw away the highest and the lowest values of the 4 to 5 measurements we read.

dirkmueller:
@ricki-z it is being shown as "-" now which means "no valid read". I am not sure it makes sense to offload filtering error codes to all the visualisation softwares.. I can see that neither opensensemap nor luftdaten maps (or my own  or madavi.de) has any kind of filtering for this. 

I think this violates the "be strict in what you produce, be relaxed in what you accept" principle.

dirkmueller:
I agree that alerting the user on failed sensors would make a lot of sense. I can work on that if you want

dirkmueller:
I did see this "remove minimal and maximal measurement, average out" kind of logic in the code path, but I'm not sure this is a great idea, as we don't know how long the tube is that the air is being sucked through. just averaging over the last 10 successful measurements is not gonna make any difference here in this case, fwiw (the readouts is consistent 999/1999 all the time). 

Also, it is documented that those values are out of range, so it makes sense to discard them. 

Regarding the logic you mentioned, I think a better solution would be to suck air until the readings are roughly stable, and then prefer the latter readings over the earlier readings, because in my setup the earlier readings are slightly more "noisy" than when the airflow has been constant over some time.

ricki-z:
We have to use a tradeoff of sensor livetime and enough measurements (to also detect short events). The SDS011 is specified for a livetime of 8000h (laser). So a longer running laser would shorten the lifetime. So we have to limit the time the SDS011 is active and can't wait until the values are "stable enough". Worst case the sensor is running without stopping and will "die" after a year.
At the moment we "prefer" the later values as we throw away the values while sensor warmup and take only the last 5 values ;-).

As most users won't look at the sensor data every day I think it's better to detect faulty sensors on the server side and alert the user.

The alerting (for different kinds of "errors") is work in progress. We need to write some kind of instructions for the users what to do.

ricki-z:
And a question: where is the invalid data shown as "-"?

dirkmueller:
okay, sorry, lets not mix the sensor lifetime discussion here. I didn't want to change the logic. 

Here's my concern. lets assume sensors are failing occassionally for some reason. I have data being sent to luftdaten, to opensensemap and to my own influxdb. I don't want erratic data in my influxdb because it screws up the rest of the logic around the data. I want it to not send any data so that it appears as "absent" (which already triggers an alarm for me). 

I understand that you do prefer to get erratic data on luftdaten. Would it be acceptable when I change the logic to filter invalid data only on targets other than luftdaten?

Marwe:
Data with such checks would be a "derived product", for now there is only the "raw data".  This way filters (e.g. for  range/gross errors) and corrections could be applied afterwards, and the derived products can themselves be developed in parallel (and updated, versioned etc.).

ricki-z:
@dirkmueller What about a checkbox to disable sending erratic readings to own APIs and InfluxDBs? So the user can decide if he wants to filter this on the sensor or would like to use it for own diagnostics?

dirkmueller:
@ricki-z works for me, I need to see how easy it will be to sense garbage to luftdaten and the filtered output to the other apis though code wise

PR #443, opened 2019-08-31
Revert reindent of cfg namespace

dirkmueller:
In f1a3583117e11581faee71b16b8b4862c721aaf6 the indentation was
changed (incorrectly) to not indent based on the opening of namespace.

Although that shouldn't matter, for some unknown reason this messes up
the travis/platformio based builds. Revert this part.


ricki-z:
This seems like a bug in astyle.

PR #444, opened 2019-08-31
debug_out() cleanups

dirkmueller:
Passing 3 parameters is code size wise an overhead over two parameters.
We can easily refactor this into:

  debug_out(..,.., 1) -> debug_outln(.., ..);
  debug_out(..,.., 0) -> debug_print(.., ..);

which are functions with the same length (so vertical aligning is the
same). this reduces size of the result by ~ 800 bytes.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1,6 +1,3 @@
-#include <Arduino.h>
-#define INTL_DE
-
 /************************************************************************
  *                                                                      *
  *  This source code needs to be compiled for the board                 *
@@ -118,6 +115,11 @@
 /*****************************************************************
  * Includes                                                      *
  *****************************************************************/
+
+#ifndef INTL_DE
+#define INTL_DE
+#endif
+
 #if defined(ESP8266)
 #include <FS.h>                     // must be first
 #include <ESP8266WiFi.h>
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -566,7 +566,7 @@ String esp_chipid;
 
 String mac_addr_sta;
 
-long last_page_load = millis();
+unsigned long last_page_load = millis();
 
 bool wificonfig_loop = false;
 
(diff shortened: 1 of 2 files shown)


ricki-z:
include paths without a leading "./" may result in usage of the wrong includes if there is a file with the same name in the path. Some it's more secure to use "./"
Naming functions: Why not naming the debug_out() replacements debug_out() and debug_outln()?
The difference between the two functions is the line break, the function names should show this (like write() and writeln()).
And a question: Do you know why the functions need less memory than one function with one more parameter?

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -604,22 +606,31 @@ const char data_first_part[] PROGMEM = "{\"software_version\": \"{v}\", \"sensor
 /*****************************************************************
  * Debug output                                                  *
  *****************************************************************/
-void debug_out(const String& text, const int level, const bool linebreak) {
+
```
ricki-z: "if 0": Really?

dirkmueller:
@ricki-z I am not sure I understand the comment about #include "./foo" vs #include "foo". both do exactly the same: look in local dir (relative to the file with the include) first, and only then traverse path. see https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html and https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

dirkmueller:
regarding function naming: I had it originally debug_out and debug_outln, as I also liked that more. the issue is that then the arguments are no longer vertically aligned. I'm okay with that personally. 

the reason this is smaller is because we don't have to push a 3rd parameter into a register/onto the stack, which requires less machine code. we simply call two different functions, and as we get rid of the if() condition then it should be marginally faster (no practical difference though) as well.

airrohr-firmware/airrohr-firmware.ino
dirkmueller: leftover. I only used this to measure the potential to spend some time on optimizing debug() statements. when removing all of them over 25kb of progmem is removed, so there is a very large potential. i've removed this now.

ricki-z:
#include "./foo" vs #include "foo"
Yes, the compiler SHOULD look first in the local directory. But the Arduino IDE seems to ignore this sometimes. We had this with the oledfonts.h and the DHT.h. In both cases the library versions were used if we haven't included "./foo.h"
So just to be sure I would like to this version.

dirkmueller:
@ricki-z I removed the include ./foo cleanup commit from the patch series. thanks for the patience.

ricki-z:
For clarification: I would like to change the includes. We used the shorter version as we switched to a 'local' copy of the DHT library. But it wasn't working with that, only with the './foo' version. And nobody knows what Arduino may change in future versions ;-)

PR #445, opened 2019-09-01
further codesize optimisations

dirkmueller:
Saves ~ 2000 bytes of flash

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -610,17 +610,15 @@ const char data_first_part[] PROGMEM = "{\"software_version\": \"{v}\", \"sensor
  * Debug output                                                  *
  *****************************************************************/
 
-void debug_out(const String& text, const int level) {
-	if (level <= cfg::debug) {
-		Serial.print(text);
-	}
-}
+#define debug_level_check if(level > cfg::debug) return;
 
-void debug_outln(const String& text, const int level) {
-	if (level <= cfg::debug) {
-		Serial.println(text);
-	}
-}
+void debug_out(const String& text, const int level) { debug_level_check; Serial.print(text); }
+void debug_out(const __FlashStringHelper* text, const int level) { debug_level_check; Serial.print(text); }
+
+void debug_outln(const String& text, const int level) { debug_level_check; Serial.println(text); }
+void debug_outln(const __FlashStringHelper* text, const int level) { debug_level_check; Serial.println(text); }
+
+#undef debug_level_check
 
 /*****************************************************************
  * display values                                                *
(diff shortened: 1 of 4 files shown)


airrohr-firmware/html-content.h
```diff
@@ -27,6 +27,7 @@ const char DBG_TXT_SDS011_VERSION_DATE[] PROGMEM = "SDS011 version date";
 const char DBG_TXT_CONNECTING_TO[] PROGMEM = "Connecting to ";
 const char DBG_TXT_FOUND[] PROGMEM = " ... found";
 const char DBG_TXT_NOT_FOUND[] PROGMEM = " ... not found";
+const char DBG_TXT_HR[] PROGMEM = "-----";
```
ricki-z: replacing strings less or equal 4 bytes doesn't save space but needs a functions call.

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -527,6 +527,15 @@ float value_SPS30_N2 = 0.0;
 float value_SPS30_N3 = 0.0;
 float value_SPS30_N4 = 0.0;
 float value_SPS30_TS = 0.0;
+
+float last_value_BMP_T = -128.0;
```
ricki-z: These lines were only moved. But now they are between vars used for PM sensors ...

airrohr-firmware/html-content.h
ricki-z: Replacing strings less or equal 4 bytes doesn't save space but needs a functions call.
In some cases you even doesn't save any space for longer strings, or you need more memory.

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -2257,7 +2257,6 @@ void wifiConfig() {
 		const IPAddress apIP(192, 168, 4, 1);
```
ricki-z: This is needed for users that forgot the password for the sensor in AP mode.
And we need this for a future function where we create a unique password at first start. So this will be the only possible way to read this password.
You need physical access to read this (debug output over USB only). So this should not be too sensitive. It's not the password for the wifi the sensor will connect to.

dirkmueller: thanks for the explanation. I've updated this commit to log this with a prefix again.

airrohr-firmware/airrohr-firmware.ino
dirkmueller: good point, fixed that. the real change is double to float though.. (to reduce float/double conversions as those values are captured as floats by the sensor libraries)

airrohr-firmware/html-content.h
dirkmueller: the assembly code *length* is exactly identical between:

```c++
debug_out("foo");
```

and 

```c++
debug_out(FPSTR("foo"));
```

thats because the former is actually implicitly doing this:

```c++
debug_out(String("foo"));
```

so we're just treating one function call over the other. in the future I'd like to add specialisations to debug_* to avoid the implicit String() construction overhead (which is both code size and runtime suboptimal). so I'll drop this again.

ricki-z: Thanks for clarification.

Issue #446, opened 2019-09-02
Build fails with error "TypeError: a2b_base64() argument 1 must be string or buffer, not None"


nospam2000:
When building with PlatformIO, the after-build step fails with the following error message:
```
Building .pio/build/nodemcuv2/firmware.bin
Creating BIN file ".pio/build/nodemcuv2/firmware.bin" using ".pio/build/nodemcuv2/firmware.elf"
after_build([".pio/build/nodemcuv2/firmware.bin"], [".pio/build/nodemcuv2/firmware.elf"])
*** [.pio/build/nodemcuv2/firmware.bin] TypeError : a2b_base64() argument 1 must be string or buffer, not None
Traceback (most recent call last):
  File "/Users/michael/.platformio/packages/tool-scons/script/../engine/SCons/Action.py", line 1054, in execute
    result = self.execfunction(target=target, source=rsources, env=env)
  File "/Users/michael/Downloads/sensors-software/airrohr-firmware/platformio_script.py", line 17, in after_build
    target_name = b64decode(ARGUMENTS.get("LANG"))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/base64.py", line 75, in b64decode
    return binascii.a2b_base64(s)
TypeError: a2b_base64() argument 1 must be string or buffer, not None

```

Before that error the following related warnings are shown in the console:
```
> Executing task: platformio run --environment nodemcuv2 <
Warning! Ignore unknown configuration option `lang` in section [env:nodemcuv2]
Warning! Ignore unknown configuration option `lang` in section [env:nodemcuv2_bg]
Warning! Ignore unknown configuration option `lang` in section [env:nodemcuv2_cz]
[...] // for all languages

```

See also the complete build log:
[build.log](https://github.com/opendata-stuttgart/sensors-software/files/3564932/build.log)

The script platformio_script.py uses `ARGUMENTS.get("LANG")` to get the command-line parameter "LANG", but "ARGUMENTS" doesn't contain "LANG". The content of ARGUMENTS is:
[...]

PR #447, opened 2019-09-02
fixed issue #446

nospam2000:
read "LANG" setting from platformio.ini via configparser instead of ARGUMENTS.get("LANG") because ARGUMENTS.get("LANG") is not defined in my recent installation.


ricki-z:
Could you please push this to the beta branch? This is where the development is done.

nospam2000:
already merged to beta branch

PR #450, opened 2019-09-03
Remove redundant initalisation of String()

dirkmueller:
This just hurts my eyes, the default String() constructor
is the same. code wise identical.


dirkmueller:
(for consideration)

ricki-z:
These are remains needed in older versions of the ESP8266 Arduino extension. These versions had some "non-standard" implementations of C functions.

PR #457, opened 2019-09-05
Sync airrohr-update-loader with current state of airrohr-firmware

dirkmueller:
Various cleanups. Add it to travis to ensure that we have
tests for it as well. Port to ArduinoJson 6.x. do debugcleaups.

Build is around 305kb for ESP8266

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ install:
 
 script:
     - cd airrohr-firmware && pwd && ls -l && platformio run && ls -l builds/
+    - cd ../airrohr-update-loader && pwd && ls -l && platformio run && ls -l builds/
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -3611,42 +3611,42 @@ String sensorGPS() {
  * AutoUpdate                                                    *
  *****************************************************************/
 static void autoUpdate() {
-	if (cfg::auto_update) {
-		debug_outln(F("Starting OTA update ..."), DEBUG_MIN_INFO);
-		debug_out(F("NodeMCU firmware : "), DEBUG_MIN_INFO);
-		debug_outln(SOFTWARE_VERSION, DEBUG_MIN_INFO);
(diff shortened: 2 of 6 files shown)


ricki-z:
Thanks for the cleanup. This is a very fast written first try for this. Only to get the size of this.
We need to check the size if the cert validation, BearSSL and HTTPS OTA is added (needs version  > 2.5 of the Arduino core for ESP8266).

dirkmueller:
@ricki-z I have switched to Arduinocore 2.5 for update-loader now. size is still 305kb. 

I am still struggling to see how I can test this OTA path on ESP32 (where the main sketch is currently for reasons that I don't understand just below 1MB). Will look at that. or buy an ESP8266 for development.

I was wondering how to derisk this a bit more. As we have plenty of space in SPIFFS, I am wondering if the main firmware shouldn't do the new fw download and save it in SPIFFS (including various validation steps). Then the updater-firmware would be only streaming this from SPIFFS into the OTA region and reboot again. This way a wifi or any other failure would not leave a sensor in a feature-crippled updater firmware with difficult recovery (no webserver interface for reboot or the like). for me that sounds like a safer path forward. 

There seems to be a streaming interface for ArduinoOTA so one could build something using SPIFFS backed file as input. of course we're copying the main fw around multiple times (first to SPIFFS, then from SPIFFS into the OTA region, then eboot from OTA to main sketch region), but I think that doesn't matter.  

Given that SPIFFS is so large, we could even store the original version FW in there and have some logic to detect that the new one didn't boot and fail back to the old one.

ricki-z:
For the ESP8266 we are limited to the first megabyte. The small routine that will move the new firmware to the start of flash memory can only access the flash up to this address. And it's not able to access SPIFFS.
The only way to change this would be an extension of the Arduino core.

At the moment we should disable OTA for ESP32. I haven't implemented the server side part for this yet.

dirkmueller:
I understand that, there are two copies involved, and eboot can only copy from the tip of the sketch area. But I think you can implement an ota procedure using the streaming interface which will stream the firmware from the spiffs instead of http(s).

Let me make a prototype.

PR #458, opened 2019-09-06
Two Stage OTA update from SPIFFS

dirkmueller:
This loader variant looks for a file called firmware.bin
on the SPIFFS and triggers a Sketch Update precedure with it.
By uploading it to the OTA flash area and configuring eboot
to copy on boot, we're in-place replacing the updater-firmware
with the final firmware.

This produces a relatively small sketch of 271kb. the main airrohr-firmware
would have to download the new firmware version, save it on SPIFFS
 and then install the airror-update-loader via OTA. This one will
then do a local Update of the firmware from SPIFFS and rename it to
an old binary.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -3425,26 +3425,125 @@ static void fetchSensorGPS(String& s) {
 	debug_outln_verbose(FPSTR(DBG_TXT_END_READING), "GPS");
 }
 
+
 /*****************************************************************
  * AutoUpdate                                                    *
  *****************************************************************/
-static void autoUpdate() {
+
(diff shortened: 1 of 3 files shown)


dirkmueller:
@ricki-z This is what I meant. I've tested this successfully on an ESP8266, by manually uploading a firmware.bin on SPIFFS and then installing this sketch. Seems to work fine and has no network dependencies. 

What is missing is a bit of validation (a verify-read..).

ricki-z:
Looks good.
So if we could be sure to have loaded the right firmware image and the right updater we could switch to this.
The updates would be a 'static file' (not changing with new firmware versions). So for this we could use a hardcoded (md5) checksum.
The firmware files are more difficult.

PR #460, opened 2019-09-06
Remove unnecessary sprintf() / static char usage

dirkmueller:
Only usage of sprintf() in this file, which is unnecessary
as it is just mere string concatenation which we can do
differently. Saves about 20 bytes of flash space as well
as looking much nicer.


ricki-z:
This part was added by a russian participant. AirCMS has forked our firmware long before. And this part should make our firmware compatible with the AirCMS servers. There are some parts in that code that could be optimized ;-)
I.e. the function hmac1 could be replaced as it is 'only': hmac1 = sha1hex(secret + sha1hex(string)), where sha1hex is only an alias for sha1 for the ESP8266, but a little more code for the ESP32.
The ESP32 part was only a quick 'solution', may be there is a more elegant way to write this.

PR #461, opened 2019-09-07
Debug code cleanup

dirkmueller:
The "below 499999 bytes program size" PR.


dirkmueller:
@ricki-z changed the naming as suggested. we can always readd max_info, it is still in the numbering scheme.

PR #462, opened 2019-09-08
Beta

nospam2000:
ESP32 related fixes


airrohr-firmware/airrohr-firmware.ino
```diff
@@ -4240,9 +4244,13 @@ static bool acquireNetworkTime() {
...
 	configTime(0, 0, WiFi.gatewayIP().toString().c_str());
 	while (retryCount++ < 20) {
-		// later than 2000/01/01:00:00:00
+		now = time(nullptr);
+#if defined(ESP8266)
 		if (sntp_time_is_set) {
-			now = time(nullptr);
+#endif
+#if defined(ESP32)
+		if (now > 17897*24*60*60) { // later than 2019/01/01:00:00:00
```
dirkmueller: doesn't that mean that sntp_time_is_set should not be true?

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -4225,11 +4225,15 @@ static bool acquireNetworkTime() {
...
+	configTime(0, 0, "pool.ntp.org"); // using UTC as timezone
 	while (retryCount++ < 20) {
-		// later than 2000/01/01:00:00:00
+		now = time(nullptr);		
+#if defined(ESP8266)
 		if (sntp_time_is_set) {
-			now = time(nullptr);
+#endif
+#if defined(ESP32)
+		if (now > 17897*24*60*60) { // later than 2019/01/01:00:00:00
```
dirkmueller: why is sntp_time_is_set now only used for ESP8266?

nospam2000: because settimeofday_cb() is not available for ESP32. The examples are using sntp_set_time_sync_notification_cb() but this is not yet available in the newest esp32platform delivered with platformio.

airrohr-firmware/airrohr-firmware.ino
nospam2000: I'm not sure if I understood your questions correctly.
sntp_time_is_set is initialized with false and only set in the callback time_is_set().
When the callback is called in the first try with "ntp.org", the second part (the change from line 4247 to 4252) will never be reached because the function returns.

dirkmueller:
@nospam2000 hah, we were looking at the same problem! I opened #463 for this, it reduces the code bloat a bit as well.

nospam2000:
> we were looking at the same problem!

@dirkmueller how shall we synchronize to avoid duplicate work? Shall we use github issues or the ESP32 chat on luftdaten.info?

airrohr-firmware/airrohr-firmware.ino
dirkmueller: Right, I agree with your assessment, sorry, didn't read the code closely. I wonder if we can remove all of the ESP conditionals in the code and use both code paths the same way?

dirkmueller:
@nospam2000 yes, we should coordinate. can you point me to the place where to find information about this "chat" on luftdaten.info ?

airrohr-firmware/airrohr-firmware.ino
nospam2000: Yes, in this specific case the ESP32 path will work for both. The error checking of the ESP8266 is more strict. The fallback in the ESP32 case to use the local router will not be triggered if the RTC has a date after 2019-01-01 and this will happen on a reboot.

That means:

- when ntp.org cannot be reached on the first boot it will fallback to the local router
- on the second boot the RTC will already have a valid date and don't do a fallback to the local router

To solve that, I would remove the fallback and replace it by the following solution:
Set multiple servers with one call:
1. ntp.org
2. the ntp server(s) announced by dhcp
3. the local router (actually this shall not be necessary because in many cases this is already covered by 2.)

roughly the code would be the following, but I need to check how to mix dhcp and fixed servers:
```
sntp_servermode_dhcp(1);
sntp_setservername(0, "pool.ntp.org");
sntp_setserver(1, WiFi.gatewayIP());

```

airrohr-firmware/airrohr-firmware.ino
dirkmueller: Agreed, thanks

airrohr-firmware/airrohr-firmware.ino
dirkmueller: I would try local first before going to the internet.. I implemented router/pool in my pr that I just opened.

Searching for ntp information in dhcp does sound like a good idea though.

nospam2000:
I was invited to the chat by (martin ät luftdaten Punkt info), but I think I can also invite you, if you give me your email address. My address is (nospam2000 ät gmx Punkt de).

airrohr-firmware/platformio.ini
```diff
@@ -13,42 +13,89 @@ src_dir = .
...
+lib_deps_esp32 = ${common.lib_deps_esp32_platform} ${common.lib_deps_esp32_external} ${common.lib_deps_generic_external}
+
 extra_scripts = platformio_script.py
 # This release is reflecting the Arduino Core 2.4.2 release
 # When the requirement for Arduino Core is raised, this
 # needs to be adjusted to the matching version from
 # https://github.com/platformio/platform-espressif8266/releases
-platform_version = espressif8266@1.8.0
+;platform_version = espressif8266@1.8.0
+platform_version = espressif8266@2.2.3 ; using Arduino core 2.5.2
```
dirkmueller: please update the comment above - but we can't upgrade to newer arduino core until OTA problem is solved.

nospam2000: already reverted. I have to switch back and forth all the time...

I'm not sure about using which SoftwareSerial variant and version.

bertrik: What OTA problem?

nospam2000: firmware size>512kB cannot be flashed using OTA:
4 MB Flash 3MB SPIFFS, 512 kB running firmware, 512kB new firmware

nospam2000: by the way: what is the real use of mDNS? Removing it would free some flash and the old version is pretty buggy

dirkmueller: I'm working on a two-stage OTA path that will allow to OTA firmwares >= 512kb. but it is halted by all the other problems in the code base, I can not well test my implementation right now, so I'm working on the other problems first. 

I believe the usecase for mdns is to export the $hostname.local as MDNS entry so that you can e.g. access http://esp8266-foobar.local/ and get to the IP address.. this is interesting in case of us providing https at some point (where certificates on hostnames are much nicer than on IP addresses). 

I personally would also like to get rid of the DNSServer dependency, it is fairly large and only needed for the initial setup (in STA_AP mode).

dirkmueller: but if we complain about size, I would just postpone the addition of the SPS30 sensor until we have solved the size problems.. this way we could relase OTAs with all the bugfixes and move forward.

nospam2000: to use "http://esp8266-<ID>.local/" you need to know the <ID> (part of the MAC address) before you connect to it. Some kind of catch22.
That's why I think it is not really useful.

nospam2000: > I would just postpone the addition of the SPS30 sensor

Good idea.

I also have a implemented a new sensor (MHZ19 CO2 sensor) but I didn't dare to create a pull-request, yet. Maybe I will do it in the next days just to store it somewhere.

bertrik: I often use it to quickly see the current PM value of my own sensor.

mDNS assigns a name to the node, but also adverttises the services running on the node (e.g. webserver, even the Arduino OTA port if you have local OTA enabled). On Android you can see this with https://play.google.com/store/apps/details?id=com.druk.servicebrowser for example, on linux you can see local services with avahi-discover,

nospam2000:
The build of the ESP32 target didn't succeed, probably because PlatformIO is too old.
Is it possible to switch to a newer PlatformIO version when the esp8266 platform version is now fixed at 1.8.0?

dirkmueller:
Yes, we can remove the cap on platformio to go with a newer version

airrohr-firmware/platformio.ini
dirkmueller: @nospam2000 the name can be discovered in the initial setup step where you connect to it in ap mode. Also I thought you can configure it..

ricki-z: mDNS is used to discover available dust sensors with the firmware flashing tool (https://github.com/opendata-stuttgart/airrohr-firmware-flasher). So usersdon't need to know how to look up the IP address (this project is also for users without much IT knowlegde). 
DNSServer is needed for the "login" feature while setting up the sensor. Most OSes send a DNS request for special host names to test internet connectivity. These requests need to be answered.

nospam2000: The description of the firmware flasher says _"connect the ESP8266 with a USB cable to your computer (micro-USB)"_. Can it really also be used for OTA flashing as a **local** OTA update? Actually that is what I was looking for because the USB upload is a pain.

ricki-z: Not at the moment. Maybe we can implement something similar with the update version @dirkmueller is working on (load new firmware to SPIFFS, update actual firmware to a small binary that will then only copy the new version from SPIFFS to flash)

ricki-z:
@nospam2000 could you please rebase your commit? There were some changes by an earlier commit that have some conflicts with your changes.

nospam2000:
Yes I already rebased, but the original beta branch and therefore now also my branch doesn't compile any longer with ESP32 with the following error (for every use of the debug functions):
```
Compiling .pio/build/heltec_wifi_lora_32_V2/src/airrohr-firmware.ino.cpp.o
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino: In function 'String SDS_version_date()':
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino:848:84: error: no matching function for call to 'debug_outln_verbose(const char*, const char*)'
  debug_outln_verbose(FPSTR(DBG_TXT_END_READING), FPSTR(DBG_TXT_SDS011_VERSION_DATE));
                                                                                    ^
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino:619:13: note: candidate: void debug_outln_verbose(const __FlashStringHelper*)
 static void debug_outln_verbose(const __FlashStringHelper* text) {
             ^
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino:619:13: note:   candidate expects 1 argument, 2 provided
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino:629:13: note: candidate: void debug_outln_verbose(const __FlashStringHelper*, const String&)
 static void debug_outln_verbose(const __FlashStringHelper* text, const String& option) {
             ^
/Users/michael/Documents/Arduino/sensors-software/airrohr-firmware/airrohr-firmware.ino:629:13: note:   no known conversion for argument 1 from 'const char*' to 'const __FlashStringHelper*'

```

nospam2000:
> the upload and monitor ports are not the mentioned on all systems. They are different between OSes

I will remove them.

The main problem is, that platformio has all configuration in one file: site/user specific and compile/linker settings. 
At the moment I have to switch back and forth before every check-in to keep that config clear from local settings. Maybe I can use environment variables to overcome this, but I think there is also no #if mechanism in platformio.ini.

nospam2000:
> Yes I already rebased, but the original beta branch and therefore now also my branch doesn't compile any longer with ESP32

The reason is this fix https://github.com/espressif/arduino-esp32/issues/1371

ricki-z:
By removing 'common includes' like <Adafruit_BMP085.h> the source code can't be compiled with the Arduino IDE. 
The directory .vscode should be included in .gitignore.

ricki-z:
Includes: Okay I have found the second block (ex-ESP32). The source code folding here on Github is sometimes 'suboptimal'.

PR #463, opened 2019-09-08
Fix config write

dirkmueller:
Commit 9e911cd broke saving
of configuration files, as now every datatype is saved as a
string, which in the case of bool is considered true even
if the string is "false". With that changing of configuration
was completely messed up.

Originally this code was using ArduinoJSON for serialization,
which was removed due to memory pressure concerns. This can
be fixed by using the DynamicJsonDocument from ArduinoJson 6.x.

This does not only fix the bug but also reduces code bloat
by 160 bytes. The memory pressure problem will be fixed in
a followup commit.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -1457,9 +1457,9 @@ static void webserver_root() {
  * Webserver config: show config page                            *
  *****************************************************************/
 
-static void webserver_config_body_get(String& page_content) {
+static void webserver_config_send_body_get() {
 	using namespace cfg;
-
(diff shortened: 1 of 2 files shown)


dirkmueller:
@ricki-z I am not 100% sure I got your point about the last_page_load, but I just removed that patch from teh patch series, as the rest of the commits are much more urgent to get it. especially the code that fixes config writing and the config page rendering :-)

ricki-z:
There are some conflicts. Could you please check this.

dirkmueller:
@ricki-z yeah, that was kinda  expected after so large changes.. rebased.

PR #464, opened 2019-09-09
Time

dirkmueller:


dirkmueller:
Esp32 doesn't have the callback api so the bool is never set. We need to figure out some other way  to ensure we're in sync.

I happen to have a Fritzbox so I will test the behavior of this PR with guest wifi.

ricki-z:
The guest wifi should be set to "Mail and web access only" which is the default.

PR #467, opened 2019-09-10
Remove DNS Server capturing on wizard

dirkmueller:
The DNS capturing isn't working well with webbrowsers
that have been used before and where you try accessing a page
that had HSTS configured. Also this strips 2000 bytes of flash.

Instead we use MDNS, which allows accessing http://192.168.4.1/ or
http://airrohr.local/ (non-Android browsers).

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -118,7 +118,6 @@
 #if defined(ESP8266)
 #include <FS.h>                     // must be first
 #include <ESP8266WiFi.h>
-#include <DNSServer.h>
 #include <ESP8266WebServer.h>
 #include <ESP8266mDNS.h>
 #include <ESP8266httpUpdate.h>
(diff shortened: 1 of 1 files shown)


dirkmueller:
(for consideration) ideally we'd use https://tools.ietf.org/html/rfc7710#section-2.1 for this, but it doesn't seem to be easy to inject that dhcp option..

ricki-z:
https://de.wikipedia.org/wiki/Captive_Portal#Erkennung
This should work. The server names requested shouldn't be cached as users normally don't call these pages. Opening the config page automatically has drastically reduced support requests.

dirkmueller:
@ricki-z yeah, this will certainly break it. I agree its an unfortunate tradeoff, and maybe its not worth it for 2kb of flash. I think I need to check how well supported the dhcp option is and see if we can inject the dhcp option instead, that should be significantly less code.

dirkmueller:
not for now.

ricki-z:
The DHCP server implementation for the ESP8266 supports only some of the possible DHCP options: https://github.com/esp8266/Arduino/blob/d8531cb2c4aec39d917fed8d642afe86af59e1c6/tools/sdk/lwip/src/app/dhcpserver.c#L135
The captive portal option is not supported.

PR #468, opened 2019-09-11
Advertise in AP mode the sensor via final MDNS name

dirkmueller:
Remove
some code duplication/indentation errors as well as
ensuring that our captive DNS responses are not cached
by setting TTL to 0.

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -132,7 +132,6 @@
 #if defined(ESP8266)
 #include <FS.h>                     // must be first
 #include <ESP8266WiFi.h>
-#include <DNSServer.h>
 #include <ESP8266WebServer.h>
 #include <ESP8266mDNS.h>
 #include <ESP8266httpUpdate.h>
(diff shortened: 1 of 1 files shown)


dirkmueller:
You're correct on this, but as we install a dns server that resolves anything to the host ip, it doesn't matter. 

Another thought I just had is that we can name this already like the host will be called in STA mode, this way you might be able to easily reaccess it after reboot (same hostname)

We could even install some js that will refresh until the page until it becomes available again with the same hostname but different ip and do a 2nd stage configuration.

ricki-z:
Sorry, my fault. I had still the commit with the DNS server removal in mind.
A first step to the js refresh could be a forwarding to the "future" hostname (including the chipID) when airrohr.local is requested. For the refresh we need to test the restart time. (And a hint that this won't work on a freshly flashed system that wasn't power cycled).

ricki-z:
Do we need the MDNS in config mode if there is a DNS server? There should be only one client and this client should have access to the sensor wifi only. So a DNS request for 'airrohr.local' should be resolved to the sensor IP anyhow.

dirkmueller:
@ricki-z interestingly further testing showed that for some combinations of OS/browsers a "*.local" is never resolved via DNS, only via MDNS.

So I have updated the change to only add MDNS and not change anything else. Again the point of adding MDNS is that we can add some JS to the client side redirecting to that hostname to be able to seamlessly able resume the browsing session after wifi switchover.

PR #477, opened 2019-09-15
Pick up new Adafruit releases for Bosch Environmental Sensors

dirkmueller:
Those include a cleanup from me that reduces memory consumption.

### DIFF
--- a/airrohr-firmware/Readme.md
+++ b/airrohr-firmware/Readme.md
@@ -83,11 +83,11 @@ In ESP8266 für Arduino IDE enthalten:
 * SoftwareSerial (GNU Lesser Public License >=2.1). Don't install additional library!!!
 
 Installierbar über Arduino IDE (Menü Sketch -> Bibliothek einbinden -> Bibliotheken verwalten, in Klammern die getestete Version und die Art der Lizenz):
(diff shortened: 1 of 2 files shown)


ricki-z:
Would be nice to have one library per manufacturer (Bosch, Sensirion) as many functions are mostly the same per library. At the moment we need 3 libraries for Bosch sensors. The Sensirion temperature sensors (SHTxx) aren't included yet because we don't had enough memory for the needed libraries ...

dirkmueller:
@ricki-z agree, I had a short look at unifying the Bosch sensors, and there are currently easier wins (especially as I don't have all those sensors around for testing)

dirkmueller:
looking at it:

```bash
xtensa-lx106-elf-nm -C -S --size-sort -t d .pioenvs/nodemcuv2/firmware.elf  | grep Adafruit_BM.280 | cut -d' ' -f2 | paste -sd+ - | bc
4788
```

so we could save maybe 2kb.. hmm.

ricki-z:
And it's the same with the Sensirion SHTxx sensors. There are 3 generations now. That was the reason not to add them to our firmware, even if some users would like to use them or use a modified firmware.
I will check the size of those libs in the next days.

PR #479, opened 2019-09-15
Merge Adafruit's BMP280/BME280 sensor readings into airrohr-firmware

dirkmueller:
BMP280 is downwards register compatible to BME280, which features
additionally a humidity sensor. By linking both adafruit libraries
and instanciating them twice we add quite a bit of overhead. Instead
we can differenciate the sensors by the sensor id (which is 0x60 for
BME280 and 0x58 for BMP280) and treat them equivalent.

This is a fork of Adafruits libraries to remove the SPI support which
we so far have not enabled.
    
Saves 4900 bytes of flash and 100 bytes of RAM.


bertrik:
Seems to work here with a BMP280, I checked out your branch, compiled and flashed the code (no SDS011 connected though)

dirkmueller:
@bertrik looks good - are the values reasonable though? 27° and 1017hPa sounds pretty high (unless you live near sealevel)

PR #483, opened 2019-09-18
[WIP] Suspend wifi radio modem between measurements

dirkmueller:
This reduces energy consumption by about 50% and the
SBC is no longer feeling "warm" when touching it, which
might reduce heating effects on the connected sensors
somewhat.

- [ ] Tested successfully


dirkmueller:
(do not merge, still not working..)

Issue #375, opened 2019-02-02
Support for MH-Z19


misev:
How hard would it be to add support for MH-Z19 (CO2 sensor)?
I've found a few seemingly related resources:
- https://github.com/popstas/nodemcu-co2-temp-hum-grafana
- https://diyprojects.io/publish-co2-concentration-mh-z19-sensor-domoticz-arduino-code-compatible-esp32-esp8266/

Unfortunately I have no experience with arduino / embedded at all, no idea where to start in order to extend the firmware in this repository.

DeeKey:
According to the datasheet this is  internal sensor.
Its working temperature is 0 ~ 50 °C
What will it show in a negative temperatures?
Also there is a note "8.1 Do not use the sensor in the high dusty environment for long time."

Though the lifespan is quite long - over 5 years!

As to my mind some other sensor must be taken into consideration.
https://www.winsen-sensor.com/sensor/carbon-dioxide-sensor.html

jordanc365:
Hi @misev I'm also very interested in attatching gas sensors to the units!
I was thinking more along the lines of something like this though as it only requires reading voltages.
[link](http://www.alphasense.com/index.php/products/carbon-monoxide-air/)

nospam2000:
> How hard would it be to add support for MH-Z19 (CO2 sensor)?

I already used this sensor in another project for indoor measurements.

> Also there is a note "8.1 Do not use the sensor in the high dusty environment for long time."

The sensor is closed and uses optical measurement, so I would interpret that note as "_the optical windows should be free of dirt_". I don't think that this is a problem. 

This is not my top priority issue, but I might start working on it.

nospam2000:
I have added support for the MH-Z19 sensor in [my mhz19sensor branch](https://github.com/nospam2000/sensors-software/tree/mhz19sensor) (based on the beta branch). The MH-Z19 is using the same serial port as the GPS sensor, so you can either use the GPS sensor or the MH-Z19 sensor, but not both at the same time. In the webserver UI the GPS has priority, that means if you enable both sensors, the MH-Z19 will be disabled when saving the configuration.

Unfortunately my soldering iron gave up yesterday, so I can't test it myself at the moment. Any volunteers for testing?

nospam2000:
The servers api-rrd.madavi.de and api.sensor.community both give the following error messages when sending CO2 data:

```
[...]

PR #472, opened 2019-09-12
added support for MH-Z19 CO2 sensor

nospam2000:
implements issue #375

PR #485, opened 2019-09-19
added support for MHZ19 CO2 sensor

nospam2000:
resolves issue #375


nospam2000:
this pull request replaces pull request #472

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -3559,10 +3610,11 @@ static void display_values() {
 	double lat_value = -200.0;
 	double lon_value = -200.0;
 	double alt_value = -1000.0;
+	double co2_ppm_value = -1.0;
 	String gps_sensor, display_header;
 	String display_lines[3] = { "", "", ""};
 	int screen_count = 0;
-	int screens[5];
+	int screens[8];
```
dirkmueller: This is a pretty huge bugfix actually (stack based buffer overflow)..

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -4418,6 +4497,11 @@ void loop(void) {
 			data += result_GPS;
 			sum_send_time += sendLuftdaten(result_GPS, GPS_API_PIN, F("GPS"), "GPS_");
 		}
+		else if(cfg::mhz19_read)
+		{
+			data += result_MHZ19;
+			sum_send_time += sendLuftdaten(result_MHZ19, GPS_API_PIN, F("MHZ19"), "MHZ19_");
```
dirkmueller: Copy and paste error here - the gps api pin needs to be changed to something mhz19 specific

airrohr-firmware/mhz19.cpp
```diff
@@ -0,0 +1,150 @@
...
+	m_pSerialPort(&serialPort),
+	ppmMaxValue(0xffff)
+{
+}
+
+/*
+set the measuring range to 1000, 2000, 3000 or 5000
+ppm range for sensor MH-Z19 is set to 2000 when sensor is delivered
+*/
+bool CMHZ19Sensor::SetRange(uint16_t range)
```
dirkmueller: Borderline nit-pick here: I see most of the rest of the code to use functions with the lower camelcase schema, aka this could be setRange() to be more consistent

airrohr-firmware/mhz19.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    return true;
+  }
+  else
+  {
+    return false;
+  }
+}
+
+uint16_t CMHZ19Sensor::GetRange()
```
dirkmueller: The whole getrange logic doesn't seem to be used? Is this expected?

airrohr-firmware/mhz19.cpp
```diff
@@ -0,0 +1,150 @@
...
+  {
+    uint8_t recvCnt = m_pSerialPort->readBytes((char*)response, sizeof(response));
+
+#if DEBUG_OUTPUT    
+    for(int8_t i = 0; i < min(recvCnt, packetLen); i++)
+    {
+      Serial.print(response[i], HEX);
+      Serial.write(' ');
+    }
+#endif
```
dirkmueller: I believe some of these lines have trailing whitespace (it somehow isn't visible in githubs view)

airrohr-firmware/mhz19.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+  const uint16_t millisecondsPerChar = 1.0/(mh_z19_baudrate / 10.0) * 1000.0 + 1.1; // 9600 Baud, 8N1
+  const uint16_t millisecondsPerPacket = packetLen * 1.0/(mh_z19_baudrate / 10.0) * 1000.0 + 10.1;// 9600 Baud, 8N1
+  delay(millisecondsPerPacket); // allow buffer to fill up
+  for(int8_t i = 0; i < (4*packetLen); i++)
+  {
+    int16_t c = m_pSerialPort->peek();
+    if(c == 0xff)
+    {
+      return true; // now we are 'probably' synchronized
```
dirkmueller: Do we want to check whether tje next byte is 0x86 ? Or whether the 8th byte matches the checksum ?

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -1035,6 +1045,9 @@ static void readConfig() {
 			strcpy(cfg::host_influx, "");
 			cfg::send2influx = false;
 		}
+		if(cfg::gps_read && cfg::mhz19_read) { // cannot be active at the same time
+			cfg::mhz19_read = false;
```
dirkmueller: I wonder why we give preference to gps.. or why there is gps support at all.. seems pretty unuseful for a stationary system

airrohr-firmware/airrohr-firmware.ino
nospam2000: I haven't checked if this is a real problem, that means I didn't count how many options can be active at the same time. I just counted the number of `screen_count++` and used that as array size to be on the safe side.

For example gps and mhz19 use the same serial interface **at the moment**, so there cannot be more than 7 pages. The problem is just that when this restriction goes away that number would also have to be changed which is not easy to detect.

If you want to save some bytes of RAM the way would be some enum or MACROS which count the maximum number  and use if/elseif for the options which cannot be active at the same time.

airrohr-firmware/airrohr-firmware.ino
nospam2000: I was using the same api pin number as the gps because it actually uses the same hardware pin.
I didn't understand the concept of the "api pin".
What's is it used for and what value shall I use?

airrohr-firmware/airrohr-firmware.ino
nospam2000: > I wonder why we give preference to gps..

This was just to not change the current behaviour.

> or why there is gps support at all.. seems pretty unuseful for a stationary system

There are some units which are not stationary and use a gps.

airrohr-firmware/airrohr-firmware.ino
nospam2000: I found [some info about the API pin](https://github.com/opendata-stuttgart/sensors-software/tree/master/airrohr-firmware#luftdateninfo-api-pins), but still not clear what it is and which number to chose

dirkmueller: I can't really tell you what the exact purpose is either..my guess is that this is some way to tag the information and match it in the server side backend. However, even id that's untrue, I would just define a new #define to the same value then instead of repeating the name here..

airrohr-firmware/airrohr-firmware.ino
dirkmueller: I think it is a real problem. If you have enough sensors it will overflow.

If we want to save space I would change that from int to uint8_t, because there is no reason for this to be full size 32bit ints..

nospam2000: I agree!

airrohr-firmware/airrohr-firmware.ino
nospam2000: > tag the information and match it in the server side backend.

yes, it is only used in sendData():
```
	request_head += F("X-PIN: ");
	request_head += pin;

```

I will define a new number 16 (one after DNMS)

ricki-z: Please use 17 as pin.
The esp8266 chipID together wuth this pin is used to identify the component for which the data is sent. We try to use the same pin for sensors with the same purpose. So all PM sensors use pin 1, the temperature/pressure sensors pin 3, temperature/humidity pin 7, temperature/humidity/pressure pin 11, GPS pin 9 (more in ext_def.h)
Some users might modify the firmware to use 2 sensors of the same type (i.e. temperature/humidity inside or outside the casing).  For such a device we could add a sensor with pin 7 and one with pin 8.

airrohr-firmware/mhz19.cpp
nospam2000: I first thought it would have an effect on the accuracy, but the only reference in the docu is for the scaling of the PWM value, not for the UART output.

Removed range logic (hidden behind #ifdef)

airrohr-firmware/mhz19.cpp
nospam2000: removed all whitespaces at EOL from all source files

airrohr-firmware/mhz19.cpp
nospam2000: No, that logic is done in readCO2Sensor(). synchronizeStream() only eats up any dangling bytes until the Start Of Frame is reached.

airrohr-firmware/mhz19.cpp
nospam2000: now using camelCasing

airrohr-firmware/mhz19.h
```diff
@@ -17,6 +17,7 @@
  ***********************************************************************/
 
 #include <Arduino.h>
+#define MHZ19_SUPPORT_RANGE
```
dirkmueller: isn't that supposed to be

#undef MHZ19_SUPPORT_RANGE 

?

airrohr-firmware/airrohr-firmware.ino
dirkmueller: submitted as https://github.com/opendata-stuttgart/sensors-software/pull/486

airrohr-firmware/mhz19.h
nospam2000: it should have been commented, but I compared the code size between the commented and uncommented version and forgot to change it back...

nevertheless in the next commit eda92fc97b06e5c65c9d69e6ad1cc46096ab22b7 it is already correct.

PR #487, opened 2019-09-21
Remove redundant parentheses around case statements

dirkmueller:
From the looks-confusing-to-me-dept.


ricki-z:
Too many programming languages at the same time ...

PR #490, opened 2019-09-22
Size reduce the Luftdaten logo

dirkmueller:
Instead of duplicating this in every intl* file we can include
a common logo definition, as it seems to be the same everwhere
except for BG.
    
Also convert this to PNG as the SVG files are fairly bloated.
In order to reduce the sizes of PNGs, we need to switch to 4bit
color mode, which is just a minor color shift (imho).
    
This reduces BG firmware by 2.5kb flash and others by about 620 bytes.


ricki-z:
Having this in intl_xx.h had the advantage that 'organizations' rolling out our sensor nationwide (i.e. Bulgaria or Sweden) could use their own logo.

dirkmueller:
@ricki-z I cannot test ios. I can confirm it shows in safari / desktop on macos. I would hope that is some indication it would work on ios as well..

dirkmueller:
@ricki-z I haven't noticed localized logos.. where was that done? The only ône not identical was ibto_bg.h but that looked like a big than a feature.

ricki-z:
You can see the Bulgarian logo at https://airbg.info/ . Sweden decided to go with our logo.

dirkmueller:
I will change this back then later today.

ricki-z:
Would it be possible to define a default logo in html-content.h and re-define it in intl_xx.h if needed?

dirkmueller:
@ricki-z changed it back to be localized, using an include file. also found that PNGs are more portable and much smaller, so switched to that instead.

PR #500, opened 2019-09-25
Reduce memory consumption of sendData

dirkmueller:
sendData() had two deep-copy capturing closures, which are
keeping full copies of data and request_headers permanently in memory,
which is extremely wasteful and triggering heap fragmentation.
    
We can simplify the code a bit and also avoid printing the http
reply by default, which removes ~ 200ms request time (waiting for serial).

### DIFF
--- a/airrohr-firmware/Versions.md
+++ b/airrohr-firmware/Versions.md
@@ -11,6 +11,7 @@ NRZ-2019-124-B6
 * BMP280/BME280 sensors are merged into one implementation
 * Switch OTA updater two a two-Stage implemenation allowing more than 512kb of sketch size
 * Fix memory corruption when using displays
+* Reduce memory consumption in data sending which allows keeping the HTTP server available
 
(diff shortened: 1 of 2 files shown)


ricki-z:
With the space now available we could try to add cert verification. All our servers use Let'sEncrypt certs.  But for this we would need to change to BearSSL.

nospam2000:
> With the space now available we could try to add cert verification.

I would understand that for private data+servers, but what is the benefit of encryption and certificate validation of data which is nevertheless published and available to everyone?

Do you mean server certificate validation or client certificate validation?

ricki-z:
Server cert. Mainly for firmware update. And I think some people in "other" countries would also prefer to send their data only to our servers.

dirkmueller:
I was experimenting a bit with verification, but I would prefer to do that after the arduino core update, as there are some improvements in that area there.

PR #501, opened 2019-09-26
String reallocation optimisations

dirkmueller:
When instrumenting the String::changeBuffer() method it becomes
clear that several codepaths are excessively reallocating strings
over time in 16 byte intervals, which causes a very fragmented
heap.

Following the guidlines in https://cpp4arduino.com/2018/11/21/eight-tips-to-use-the-string-class-efficiently.html
we can group frequent string concatenation result strings into
larger buckets, preallocate memory and be more efficient in runtime
and heap usage.

* Also remove Float2String(), does the same like String() just slow (with additional spacing which gets trim'ed).

* Add more pagination to config_body_get()

* Remove 3 parameter variant tmpl (only used once, easy to replace inline)

* Change all webserver pages to be rendered incrementally, which simplifies
code paths somewhat and reduces memory pressure. Also, send_P directly
larger html header parts that don't need to be modified.

* Use CPP string concatenation instead of String.replace() at runtime where
feasible.

Overall this reduces also codesize by ~ 2kb.


ricki-z:
float2string was added at a time where this wasn't available in the Arduino core for ESP8266. And as it did what we expected we haven't replaced it.

ricki-z:
And Float2String is limiting the output to 2 or n decimal places! This is needed for i.e. for display output. String doesn't do that.

dirkmueller:
@ricki-z I see no difference in my testing of the code. the Arduino implementation is this:

```
String::String(double value, unsigned char decimalPlaces) {
    init();
    char buf[33];
    *this = dtostrf(value, (decimalPlaces + 2), decimalPlaces, buf);
}
```

decimalPlaces defaults to 2. the airrohr implementation did this:

```
// Convert a float to String with two decimals.
	char temp[15];

	dtostrf(value, 13, digits, temp);
	String s = temp;
	s.trim();
	return s;
```

so the only difference is hardcoded 13 vs "decimalPlaces +2". which seems to be doing the same thing.. 

can you show me the difference?

ricki-z:
I didn't know that this implementation has now a default of 2 decimal places. And the first occurences of String() I had found were without decimal places. So for me I looked like the complete float or double were printed (and what I would expect for String()).

PR #511, opened 2019-10-04
Don't hang on reading reply when sending data to server

dirkmueller:
If replies are not coming, time out quickly. Also don't try to read
more than is currently available to not block unplanned.


dirkmueller:
this is already in #506

PR #512, opened 2019-10-07
Updated missing translations in firmware

DavidMoraisFerreira:
Hi,
I translated some missing strings in French, Portuguese, Luxembourgish and a couple in Spanish.


ricki-z:
Thank you for the translations. They were included in the actual beta version and will be part of the next release version.

Issue #513, opened 2019-10-09
NRZ-2019-124-B7: Adafruit Si7021 / HTU21D stopped working after update to this version


zdroyer:
After the last beta update on October 1st, 2019, my temp+hum. sensor Si7021 (https://www.adafruit.com/product/3251 , HTU21D compatible),  stopped working.

I changed the sensor to a no-name clone and it worked (and still is working). 
But I tested the "faulty" sensor with another NodeMCU, and it seems that the 2019-124-B7 firmware is "faulty": when I switched back to the stable firmware (in the config), the Adafruit sensor works properly. I switched the firmware stable<->beta7 several times and the result is always the same: stable - works, beta - doesn't.

It seems the htu21d.begin method of the sensor fails (line 3937). 
Here is the serial debug output:
```Airrohr: NRZ-2019-124-B7
mounting FS...
opened config file...
parsed json...
output debug text to displays...
6
Connecting to ****
..............
WiFi connected
IP address: 192.168.43.46
Setting time using SNTP
Thu Jan  1 08:00:07 1970

..........
NTP time  ... not found
twoStageAutoUpdate
Success downloading: /airrohr/beta/latest_en.bin.md5
NewFW md5: 90b0c22407b1ff707a94fadd7e0206a3
Sketch md5    : 90b0c22407b1ff707a94fadd7e0206a3
No newer version available.
Starting Webserver... 192.168.43.46

ChipId: *******
Read HTU21D...
Check HTU21D wiring
Send to :
luftdaten.info
Madavi.de```

ricki-z:
Could you please check the latest beta version NRZ-2019-124-B8?
We are using the Adafruit HTP21D library and I have tested the latest beta with an HTU21D. The values are read. Both APIs (Luftdaten and Madavi) show the values.

zdroyer:
B8 checked - still not working.
In the meantime I found the solution: https://github.com/opendata-stuttgart/sensors-software/pull/514

zdroyer:
The Si7021 is compatible with HTU21D in reporting the measured values, but returns 0x3A if asked for read register ( https://github.com/adafruit/Adafruit_Si7021/blob/master/Adafruit_Si7021.cpp vs. https://github.com/adafruit/Adafruit_HTU21DF_Library/blob/master/Adafruit_HTU21DF.cpp)
[...]

PR #514, opened 2019-10-10
Fix for #513 : checking presence of Si7021 in case HTU21D is not dete…

zdroyer:
…cted. Added include for axTLS.

Si7021 is compatible with HTU21D, but it returns 0x3A instead of 0x2 on I2C command 0xE7, thats why the .begin() function failed

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -105,7 +105,7 @@
  *
  ************************************************************************/
 // increment on change
-#define SOFTWARE_VERSION_STR "NRZ-2019-124-B8.1"
+#define SOFTWARE_VERSION_STR "NRZ-2019-124-B9"
 const String SOFTWARE_VERSION(SOFTWARE_VERSION_STR);
 
 /*****************************************************************
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -105,7 +105,7 @@
  *
  ************************************************************************/
 // increment on change
-#define SOFTWARE_VERSION_STR "NRZ-2019-124-B8"
+#define SOFTWARE_VERSION_STR "NRZ-2019-124-B8.1"
 const String SOFTWARE_VERSION(SOFTWARE_VERSION_STR);
 
 /*****************************************************************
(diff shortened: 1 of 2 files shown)


zdroyer:
Well, I tried to compile the firmware without the axTLS header, but in the latest library 2.5.2  the (https://github.com/esp8266/Arduino/blob/2.5.2/libraries/ESP8266WiFi/src/WiFiClientSecure.h) the #include "WiFiClientSecureAxTLS.h" is commented out because of defaulting to BearSSL.

Regarding the version string, mea culpa, correcting it now.

dirkmueller:
@zdroyer yes, the current build needs to be built against arduino core 2.4.2. the pr for 2.5.2 compatibility is #506

airrohr-firmware/airrohr-firmware.ino
```diff
@@ -3935,8 +3936,17 @@ static void powerOnTestSensors() {
 	if (cfg::htu21d_read) {
 		debug_outln_info(F("Read HTU21D..."));
 		if (!htu21d.begin()) {
-			debug_outln_error(F("Check HTU21D wiring"));
-			htu21d_init_failed = true;
+      debug_outln_info(F("Read Si7021 replacement for HTU21D ..."));
+      Wire.beginTransmission(HTU21DF_I2CADDR);
+      Wire.write(HTU21DF_READREG);
+      Wire.endTransmission();
+      Wire.requestFrom(HTU21DF_I2CADDR, 1);
```
dirkmueller: so this seems to be a ugly workaround for 
https://github.com/adafruit/Adafruit_HTU21DF_Library/blob/master/Adafruit_HTU21DF.cpp#L49

wouldn't it be better to submit this as a PR to that library? then this code needs no modification.

dirkmueller:
@zdroyer I sent an alternative PR which should address this regression you found as well. Please let me know if there are any issues with that.

zdroyer:
> @zdroyer I sent an alternative PR which should address this regression you found as well. Please let me know if there are any issues with that.
Hey Dirk
Pretty clever trick - works for me! Thanks.
I'll make a PR to the Adafruit lib and I hope the lib will be updated and your extended sensor init condition will not be needed anymore.

zdroyer:
> 
> 
> @zdroyer I sent an alternative PR which should address this regression you found as well. Please let me know if there are any issues with that.

The PR is waiting for Adafruit acceptance: https://github.com/adafruit/Adafruit_HTU21DF_Library/pull/16

PR #527, opened 2019-10-30
docu: correct filename of ext_def in description 

mrMoe:
Correct a typo in the ext_def.h filename
Add table for filename descriptions for better html rendering


dirkmueller:
@mrMoe you might want to change your wifi password now..

PR #542, opened 2019-11-06
playing with WiFI

dirkmueller:


ricki-z:
In airrohr_update_loader.ino the WiFi include is missing. Or the WiFi.forceSleepBegin() should be removed.

dirkmueller:
@ricki-z removed the related commit that fails the update-loader build..

PR #549, opened 2019-11-13
Revert reporting change

dirkmueller:


ricki-z:
We should "execute" the command, but may be with a little delay (100ms?).
Some users test their SDS011 with a python script. This script is setting the query mode (which is permanent). And with that the firmware won't read any data.

PR #554, opened 2019-11-14
Turn off modem sleep

dirkmueller:
Until we can update the Arduino core 2.6.0 (which is hopefully soon)
with the newer xtensa toolchain, it seems the best bet is to
reduce wifi sleeping. Also high transmit power can interfere with
the flash reading which can cause instability.

https://github.com/esp8266/Arduino/issues/5998#issuecomment-549068473
https://github.com/esp8266/Arduino/issues/6471

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -2316,7 +2316,9 @@ static void connectWifi() {
 #if defined(ESP8266)
 	// Enforce Rx/Tx calibration
 	system_phy_set_powerup_option(1);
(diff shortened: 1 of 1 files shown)


ricki-z:
core 2.6.1 is out

CRCinAU:
Not sure what this project is, but you probably want WIFI_MODEM_SLEEP.

WIFI_LIGHT_SLEEP is broken and disconnects all the time. WIFI_NONE_SLEEP means you use ~70mA all the time. WIFI_MODEM_SLEEP will draw ~10mA on idle and doesn't seem to drop the wifi connection.

PR #564, opened 2019-11-26
Update intl_it.h

stewecar:
Few more translations in Italian from line 98 to 102


dirkmueller:
Thanks, this was created from master branch.. I cherry-picked that and merged it into #565

PR #567, opened 2019-11-26
Update dutch (nl) translations.

bertrik:
Missing translations added, one existing translation corrected.


airrohr-firmware/intl_nl.h
```diff
@@ -104,7 +104,7 @@ const char INTL_TIME[] PROGMEM = "Tijd";
 const char INTL_SIGNAL_STRENGTH[] PROGMEM = "Signaalsterkte";
 const char INTL_SIGNAL_QUALITY[] PROGMEM = "Signaalkwaliteit";
 #define INTL_NUMBER_OF_MEASUREMENTS "Aantal metingen:"
-#define INTL_TIME_SENDING_MS ""
+#define INTL_TIME_SENDING_MS "Tijdsduur opsturen metingen"
```
dirkmueller: currently other translations end this with ":" - however I just noticed that this is inconsistent. maybe make it consistent with line 106 ?

bertrik: Yes, I think that's a good idea, I will fix that this evening.

I did notice it was a bit inconsistent, the german language file has the ":", but the english one does not. I think I'll follow the german phrases and also keep it consistent with INTL_NUMBER_OF_MEASUREMENTS

dirkmueller: Fyi I indepentently decided to drop the colon everywhere see #565. That's much nicer imho

PR #575, opened 2019-12-01
Release 126-B9 as stable

dirkmueller:


dirkmueller:
Will rebase after #576 is merged

PR #583, opened 2019-12-07
Improve English translations

liutikas:


airrohr-firmware/intl_en.h
```diff
@@ -48,15 +48,15 @@ const char INTL_LCD1602_3F[] PROGMEM = "LCD 1602 (I2C: 0x3F)";
 const char INTL_LCD2004_27[] PROGMEM = "LCD 2004 (I2C: 0x27)";
 const char INTL_DISPLAY_WIFI_INFO[] PROGMEM = "Display Wifi info";
 const char INTL_DISPLAY_DEVICE_INFO[] PROGMEM = "Display device info";
-const char INTL_DEBUG_LEVEL[] PROGMEM = "Debug&nbsp;Level";
-const char INTL_MEASUREMENT_INTERVAL[] PROGMEM = "Measuring interval";
+const char INTL_DEBUG_LEVEL[] PROGMEM = "Debug level";
```
dirkmueller: why did you remove the nonbreaking space here?

liutikas: Some of the other translations do not have it. Almost every other string does not contain it. From local testing it renders the same way without it.

liutikas: Would you like me to undo this part of the change?

dirkmueller: I looked a bit at this. the nbsp isn't strictly needed as of today because there is INTL_DURATION_ROUTER_MODE in the same table column with nbsp, and the translation of that one is almost certainly going to be the longest. 

Maybe you could fix up the 5 other cases that do not use &nbsp to make it consistent (most of them do use it).

dirkmueller: I fixed this now in #594. thanks for the contribution!

PR #650, opened 2020-01-25
Create intl_hu.h

cskolly:
Firmware Hungarian translations


ricki-z:
Please use the beta branch for your pull requests. And also the actual translation files there. The master is the actual stable version where we don't do any direct changes to.

PR #654, opened 2020-01-28
Hungarian translation

netmaniac:
Added Hungarian translation


ricki-z:
Hello,
please push this to the beta branch. The master branch isn't changed until the next stable release (which is the actual beta at that moment).
There is another hungarian translation (pull request #650) that was closed for the same reason.

PR #660, opened 2020-02-01
Update pmapp link

marcauberer:
As the developer of the pmapp, I would prefer, if we link to the new, official website of the pmapp.


ricki-z:
Will be included in the next release version and the next beta.

gearshead:
Looks good(*_*)

PR #677, opened 2020-02-21
Translation update

ddzieduch:


DavidMoraisFerreira:
@ricki-z: Seems that something went wrong when you added INTL_REPORT_ISSUE, there are now two defines for this macro name. ([L38](
https://github.com/ddzieduch/sensors-software/blob/5774d80afdb36ee781e8d6b4d974831a3aeaad14/airrohr-firmware/intl_en.h#L38) and [L84](https://github.com/ddzieduch/sensors-software/blob/5774d80afdb36ee781e8d6b4d974831a3aeaad14/airrohr-firmware/intl_en.h#L84)).

Furthermore, the translation in Spanish is not identical for both strings, and the Luxembourgian string is wrong.

ddzieduch:
@DavidMoraisFerreira You're right. I started these changes on master, where there were no translations yet. I can revert this pull request.

ricki-z:
@ddzieduch: I have reverted the merge. So please recreate your translation update based on the beta branch.

DavidMoraisFerreira:
@ricki-z: Each translation file already has `#define INTL_REPORT_ISSUE` that is valid. Therefore I don't need to commit any new translations. :)

ricki-z:
But you know what you tried to update. I have to check all language files. I need a correct pull request with only the changes you made.

PR #685, opened 2020-03-07
Use HTTPS for Sensor.Community and Madavi.de

liutikas:
Both of these services are enabled by default, thus it is important to use a secure HTTPS method of sending the data.

### DIFF
--- a/airrohr-firmware/airrohr-cfg.h
+++ b/airrohr-firmware/airrohr-cfg.h
@@ -48,9 +48,7 @@ enum ConfigShapeId {
 	Config_temp_correction,
 	Config_gps_read,
 	Config_send2dusti,
-	Config_ssl_dusti,
 	Config_send2madavi,
-	Config_ssl_madavi,
 	Config_send2sensemap,
 	Config_send2fsapp,
 	Config_send2aircms,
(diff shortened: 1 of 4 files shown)


ricki-z:
Won't work. HTTPS for all sensors would overload our servers.
And what would be the benefit? We are an open data platform.

ThomasWaldmann:
@ricki-z what is "would overload our servers"?
- A guess?
- An estimation based on data? Which?
- Experience?

There are quite a lot of busy web sites these days and most use https.

I am asking because I just implemented TLS for Multigeiger, see https://github.com/ecocurious2/MultiGeiger/pull/236 and https://github.com/ecocurious2/MultiGeiger/issues/184.

ricki-z:
Experience.
A server with 32 GB RAM and 16 cores was overloaded as we set the default for the data transmission to TLS. At this time we had only around 5000 sensors online at the same time. Now we have 11.500 devices sending data.
If you deliver content for "normal" users with smartphones or computers you can handle much more. But the NodeMCU needs much more time to do the de/encryption, holding the connections open much longer. So you have more connections in parallel. And if you look at the settings of your webserver: every connection will need some RAM.
Traffic will also grow, in our case by factor larger 10.

ThomasWaldmann:
@ricki-z guess one can either have (see our tls ticket for details):
- quick https requests (50ms) and many concurrent connections (by using "connection: keep-alive")
- slow https requests (1200ms) and fewer concurrent connections (by using "close")

So, you tried both and ran into issues with both?

ricki-z:
We tried both, but it's some time ago.
The TLS support in the ESP8266 core for Arduino is changing heavily from version to version. Maybe we should look for this with the newest one again. But testing without killing the server is difficult. For this we would need some hundred test devices ...

ThomasWaldmann:
@ricki-z i guess using https://locust.io one could simulate the client side of a lot of users / sensors.

BTW, move this rather to a ticket than hijacking this PR?

dirkmueller:
I'm relative positive that it got better with the switch to bear ssl which allows session ticket resumption, removing more than 50% of the connection setup cost of ssl. 

what we could do as a first step is enable the ssl session by default for new installations, which should fade this load wise in over having everyone switch completely immediately.

PR #686, opened 2020-03-07
Fixed bugs preventing the firmware from compiling for and running on ESP32

SG-O:


ricki-z:
Seems okay so far. But could you explain the need of (just for clarification):
serialSDS.enableIntTx(true);

SG-O:
> Seems okay so far. But could you explain the need of (just for clarification):
> serialSDS.enableIntTx(true);

The serialSDS.enableIntTx was moved into the ESP8266 block above as it isn't supported by the ESP32-arduino hardwareSerial library. It is specific to the SoftwareSerial lib.

ricki-z:
Okay, haven't seen this.

PR #702, opened 2020-03-30
Next PM Sensor added

pjgueno:
We still have to check:

- values x 10 in the local webpage

- a few debug functions


pjgueno:
A few stuffs changed. For example STOP mode removed because no pin available.

pjgueno:
I corrected the PM values display in the local webpage.
I changed a few debug and add the "state code" of the Next PM sensor. You can now verify that the sensor is in good working order.
I added the particles count for PM1, PM2.5 and PM10 because the Next PM Senor is capable of.

DeeKey:
Is this sensor already working?

pjgueno:
> Is this sensor already working?

Hi. 
Yes it's working on the dev branch of the firmware @opendata-stuttgart but the pull request was not integrated yet because we are currently working on other subjact than increasing the number of supported sensors.

Anyway I can help you if you want to use the Next PM in other applications.

Feel free to ask me.

sumerland:
Does this mean the Next-PM can now be used with the current beta (beta & OTA activated in the config)? Is this sensor put to sleep in the same way as the SDS011 in order to increase the life time? It's quite an investment but I would like to see less humidity dependency of the PM values...

dirkmueller:
@ricki-z can we please reset the history of the beta branch back to the master state? I'll volunteer to fix this up, but commits like https://github.com/opendata-stuttgart/sensors-software/pull/702/commits/12fd307d180b4b8a32d5485f7e78805c16958d2e are quite   desastreous for git commit history when trying to find out what happened why and where..

pjgueno:
@ricki-z: remove everything like suggested by @dirkmueller. Actually Tera is not responding anymore. @dirkmueller maybe can you explain me good practice on commit. It worked perfectly when I tested it with a board. And actually my add ons consist only in new variables and functions called exactly like the ones for the other sensors.

dirkmueller:
@pjgueno I do believe you that you tested *with* the PM sensor. the main issue that hit my attention is that the changes were done in a way that broke every other serial sensor other than the Next PM sensor. 

In the interest of not regressing what has worked, the current code is disabled from build, you need to set the 
```
#define NPM_READ 1
```

in either platformio.ini or in ext_def.h  and then rebuild.

PR #703, opened 2020-03-30
Corrected wrong/clumsy translations in Bulgarian language.

alex-kostadinov:
As a user from Bulgaria I spotted some translations are incorrect or a bit clumsy, so I directly fixed them.


Petarkir2000:
Did you checked the beta version?

alex-kostadinov:
I just did and there translations are already corrected. In such a case, I'll close the MR, but can you give me some info why `master` does not hold the latest changes and seemingly `beta` is the most up-to-date branch?

ricki-z:
Look at the readme in the root of this repo (last point, "Development"):
`The development is done in the beta branch.`
Master is the stable version. After publishing a new stable version this isn't changed anymore until the next stable version.
This includes also translations as we also may add or delete lines in the language files.

PR #706, opened 2020-04-12
Added a webinterface based firmware updater

SG-O:
This adds a updater that allows an user to upload an update to the sensor when automatic updates have been disabled. It works on both ESP32 and ESP8266


SG-O:
I am reworking this pull request to improve code quality. I am doing this in connection with the OTA Updater for the ESP32, which I will publish soon.

Issue #731, opened 2020-06-06
BME280 sensor does not handle humidity read errors gracefully


pillilz:
fetchSensorBMX280 handles read temperature and pressure errors gracefully, by checking with isnan(.), but fails to do the same for humidity.
As a result NaN can be sent as a humidity value to downstream receivers, rather than leaving out humidity. At least one downstream receiver, Influxdb,  can't handle NaN, resulting in an error.

PR #732, opened 2020-06-06
Fixed bug in fetchSensorBMX280

pillilz:
The issue occurs if when sensor returns NaN for humidity. This value was
handled correctly for temperature and pressure, but not for humidity.

Fixes  opendata-stuttgart/sensors-software#731


ricki-z:
Please push your commits to the beta branch. We don't change the master branch until the beta version is moved to release version.

pillilz:
Created pull request #734 for the beta branch

PR #734, opened 2020-06-08
Fixed bug in fetchSensorBMX280

pillilz:
The issue occurs if when sensor returns NaN for humidity. This value was
handled correctly for temperature and pressure, but not for humidity.

See also #731


ricki-z:
I have seen this to late and had to revert this again. The fetchSensorBMX280 also reads the BMP280 which hasn't humidity. So the BMP280 would return NaN in any case.

PR #736, opened 2020-06-10
Switch to HTTP for the fsapp endpoint

marcauberer:
Hi,
we're also slowly getting trouble with the server workload due to https encryption, so it would be better to downgrade to HTTP.
Spoiler: We currently work on a new high-performance API, which will be hopefully introduced in the near future.

Are there more changes required than changing the port to 80?

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -741,7 +741,6 @@ static void createLoggerConfigs() {
 	loggerConfigs[LoggerSensemap].destport = PORT_SENSEMAP;
 	loggerConfigs[LoggerSensemap].session = new_session();
 	loggerConfigs[LoggerFSapp].destport = PORT_FSAPP;
-	loggerConfigs[LoggerFSapp].session = new_session();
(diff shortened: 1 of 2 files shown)


marcauberer:
Could we merge this pr?
Btw: I don't know why the travis build fails. I did not change significant things. Looks like the Slovakian build does not work.

marcauberer:
Could we finally merge this?!

dirkmueller:
This patch is missing the "turning off ssl" part. now you'll simply get the clients send https on port 80, which isn't what you want. 

please change in airrohr-firmware.ino:

```C++
static void createLoggerConfigs() {
...
        loggerConfigs[LoggerFSapp].destport = PORT_FSAPP;
        loggerConfigs[LoggerFSapp].session = new_session();
```

and remove the ".session" line (this will turn off ssl).

dirkmueller (changes requested):
see earlier comment

marcauberer:
Done. Should be fine now.

dirkmueller:
Yes thank you. Now please squash the commits to one ( e.g. git rebase -i HEAD~3 ; git pushf .. )

dirkmueller (approved):

PR #755, opened 2020-08-14
Fix broken link to OneWire

liutikas:
It was linking to https://github.com/opendata-stuttgart/sensors-software/blob/master/airrohr-firmware/www.pjrc.com/teensy/td_libs_OneWire.html which is not a real page.


dirkmueller:
@ricky-z as a maintainer you can simply change the PR target by clicking on the branch name

ricki-z:
@dirkmueller I could change the PR target. But if the user is sending the next PR I have to do this again (and again and ...) So the user should change its local fork to the right target to save us some work ;-)

dirkmueller:
@ricki-z then please configure the default branch in "settings" of github to beta to avoid that hassle ;)

ricki-z:
The master branch is the default. This branch should be used by people that would like to compile the actual firmware by themself (not trusting the donwload, checking what their nodemcu is executing ...). These people should use the master branch. Developers should be able to read a readme.md or contributing.md.

dirkmueller:
That issue of users not contributors could be solved by using github releases / git tags. The builds are almost reproducible from the same git tag usually.

PR #784, opened 2020-09-07
Improve logging on GPS_datetime

dirkmueller:
The idea of serial logging was that it logs what we sent to the API.


ricki-z:
I have merged the PR og @KoffeinKaio, which should to the same

dirkmueller:
@ricki-z: not quite. I will rebase this PR and resend

KoffeinKaio:
Its the seperate Date and Time to one DateTime debug Output thats missing in my PR.

@dirkmueller: be sure not to add quote like you did in your pr, as this will break stuff again.

Issue #787, opened 2020-09-10
airrohr-firmware does not compile for esp32


liutikas:
I attempted to build airrohr-firmware for `lolin d32 pro` in arduino IDE, but I am unable compile. I get the following errors:

```
airrohr-firmware/airrohr-firmware.ino: In function 'void yield_for_serial_buffer(size_t)':
airrohr-firmware:784:13: error: 'class HardwareSerial' has no member named 'perform_work'
   serialSDS.perform_work();
             ^
airrohr-firmware/airrohr-firmware.ino: In function 'String SDS_version_date()':
airrohr-firmware:797:13: error: 'class HardwareSerial' has no member named 'perform_work'
   serialSDS.perform_work();
             ^
airrohr-firmware/airrohr-firmware.ino: In function 'void webserver_status()':
airrohr-firmware:1934:21: error: 'class EspClass' has no member named 'getFullVersion'
  versionHtml += ESP.getFullVersion();
                     ^
airrohr-firmware:1938:77: error: 'class EspClass' has no member named 'getHeapFragmentation'
  add_table_row_from_value(page_content, F("Heap Fragmentation"), String(ESP.getHeapFragmentation()), "%");
                                                                             ^
airrohr-firmware:1963:64: error: 'class EspClass' has no member named 'getResetReason'
  add_table_row_from_value(page_content, F("Reset Reason"), ESP.getResetReason());
                                                                ^
airrohr-firmware/airrohr-firmware.ino: In function 'void connectWifi()':
airrohr-firmware:2368:7: error: 'class WiFiClass' has no member named 'hostname'
  WiFi.hostname(cfg::fs_ssid);
       ^
airrohr-firmware/airrohr-firmware.ino: In function 'void setup()':
airrohr-firmware:4213:12: error: 'class HardwareSerial' has no member named 'enableIntTx'
  serialSDS.enableIntTx(true);
            ^
airrohr-firmware:4238:38: error: 'class EspClass' has no member named 'checkFlashConfig'
  if ((airrohr_selftest_failed = !ESP.checkFlashConfig(true) /* after 2.7.0 update: || !ESP.checkFlashCRC() */)) {
                                      ^
```

[...]

PR #796, opened 2020-09-17
Fixes for ESP32 targets

peterhinson:
Fixes a compile time error and runtime crash on ESP32 targets (see  #787). Adds two tested ESP32 targets.


dirkmueller (approved):
LGTM. thank you for your fixes.

Issue #789, opened 2020-09-12
Network instability with NRZ-2020-130-B8


Phaze-III:
After an OTA upgrade to NRZ-2020-130-B8 network connectivity became very unstable.

- increased number of WiFi errors (at least a few every hour, used to be practically none)
- increased number of upload errors, including to the local InfluxDB, resulting in a lot of measurements gaps
- WebUI unresponsive or slow (timeouts and empty responses) most of the time

Reverting back to the NRZ-2020-129 release firmware makes the sensor stable again. NRZ-2020-130-B6 also was very stable.

Phaze-III:
System parameters fetched every 15 minutes from status page for period upgradeing from 130-B6 to 130-B8:
![Sensor System Status - Grafana B6-B7-B8](https://user-images.githubusercontent.com/54861264/93002604-764db800-f538-11ea-8dd1-be999e1ee139.png)

System parameters for period downgrading from 130-B8 to 129-release:
![Sensor System Status - Grafana B8-REL](https://user-images.githubusercontent.com/54861264/93002671-e0665d00-f538-11ea-8d19-c95de1645a45.png)

dirkmueller:
@Phaze-III thanks for the statistics, that's very helpful. In B8 we switched to a newer version of core. However, the wifi code etc (so the SDK) remained the same version like before. So for now I would assume that there is some other problem that we can trigger elsewhere as well if we know how.

Can you describe which features you configured on your airrohr? Which sensors are enabled? Which APIs? Do you use any displays?

Is there anything noteworthy about your WiFi setup? What's the wifi quality (RSSI) you that you get?

Phaze-III:
Hi @dirkmueller, thanks for looking into this.

**Sensoren**
- SDS011 (Feinstaub)
- BMP280/BME280 (Temperatur, rel. Luftfeuchte, Luftdruck)

**APIs**
API Sensor.Community  (HTTPS)
API Madavi.de  (HTTPS)
Senden an InfluxDB  (**no** HTTPS)

No displays (the sensor is placed outside)

WiFi access point is a Cisco EPC 3928 (both 5Ghz and 2.4 Ghz, using different SSID's). The 2.4 Ghz band is rather busy (lot of neighbours) but still relatively good when using a fixed channel assignment. The AP is setup as N-only and the default 20Mhz channel width.

Quality as measured by the sensor is between 70% and 80% (-70dBm to -60dBm).

[...]

PR #798, opened 2020-09-24
Redo fetchSensorGPS() if/else logic for gps_datetime

Phaze-III:
This moves the if/else code for gps_datetime inside the gps.time.isValid() conditional.
It removes one && conditional and one else clause. The code is a bit more 'ugly' but fixes the performance problems in #789


Phaze-III:
Although this does help a bit it still doesn't solve the instability. Pull request retracted.

PR #800, opened 2020-10-02
improved code

Ajay-Kanojiya:
1.Spelling corrected URL instead of URI.
2. to print the data over the Serial monitor we have to use Serial.write() instead of Serial.println().

### DIFF
--- a/esp8266-arduino/dht22-wifi/dht22-wifi.ino
+++ b/esp8266-arduino/dht22-wifi/dht22-wifi.ino
@@ -64,7 +64,7 @@ void sendData(const String& data, int pin=-1) {
     return;
   }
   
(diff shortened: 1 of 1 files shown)


dirkmueller:
:shirt:

dirkmueller (changes requested):
* This part of the code tree isn't maintained anymore. Please find a contribution under airrohr-firmware/ instead. 

* Please don't comment out code, and please don't add your commit message as a comment inside the code. the source control system provides means for commit messages, use them. 

* Regarding URL/URI - this is borderline nitpicking, but URI is more correct. for details please see https://tools.ietf.org/html/rfc3986#section-1.1.3

PR #806, opened 2020-10-10
fix: correct pressure from pascal to hectopascal

mrMoe:
The pressure is read in pascal but should be returned in hectopascal

This conversion should be done in one single place so every output shows the same value. 
This is not the case for the webview and the prometheus endpoint.


ricki-z:
The BME280 sensors are returning pascal (as most pressure sensors do). And we are saving the values with the unit used by the sensors. Pascal is converted to hectoPascal only for the displays and the local values page.

PR #813, opened 2020-10-18
Return OpenMetrics compatible format

dirkmueller:
The prometheus endpoint can be easily converted into an OpenMetrics
endpoint by adding an "# EOF" extra line at the end, which indicates
that scraping was completed successfully.

Also optimize string allocation / reallocation a bit.

### DIFF
--- a/airrohr-firmware/Versions.md
+++ b/airrohr-firmware/Versions.md
@@ -1,3 +1,6 @@
+NRZ-2020-132-B2
+* Adjust /metrics to be OpenMetrics protocol compatible
+
 NRZ-2020-132-B1
 * next beta version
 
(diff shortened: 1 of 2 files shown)


ricki-z:
Is Prometheus able to parse this? The endpoint was added for Prometheus users.

mrueg:
If it's only adding the `# EOF` prometheus will ignore it (as of right now).

you can take a look at https://www.robustperception.io/checking-openmetrics-output-is-valid


if you currently check the metrics, this is what `promtool check metrics` will lint:

```
BMP_pressure no help text
BMP_temperature no help text
SDS_P1 no help text
SDS_P2 no help text
humidity no help text
interval no help text
last_sample_age_ms metric names should not contain abbreviated units
last_sample_age_ms no help text
max_micro no help text
min_micro no help text
number_of_measurements no help text
samples no help text
sending_intervall_ms metric names should not contain abbreviated units
sending_intervall_ms no help text
signal no help text
software_version no help text
temperature no help text
uptime_ms metric names should not contain abbreviated units
uptime_ms no help text
```

dirkmueller:
@ricki-z yes, it is fully backward compatible.

PR #826, opened 2020-11-03
Remove favicon listing from header

dirkmueller:
The listing of a favion in the header causes at least Chrome to fetch
the icon immediately, while the page is still loading and validating
CSS and other assets. This leads to parallel execution of requests,
which is hanging as ESP8266 WebServer only processes one request at a
time. This can make the webui appear very unresponsive.

Switch to an implicit favicon.ico fetch, which the browser executes
last, after finishing to load the site content (e.g. after
First-Contentful-Paint).

### DIFF
--- a/airrohr-firmware/Versions.md
+++ b/airrohr-firmware/Versions.md
@@ -1,3 +1,6 @@
+NRZ-2020-132-B3
+* Move favicon.ico fetching last
+
 NRZ-2020-132-B2
 * Adjust /metrics to be OpenMetrics protocol compatible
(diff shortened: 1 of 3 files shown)


bertrik:
Hi, I introduced the favicon. it seemed like a totally harmless thing to do at the time, because it required only some html, not actual code. Good to see that you identified this as a problem for stability, I can imagine that it took some time to discover.

DeeKey:
Wow! Great shot!!!! Thank you Dirk!

PR #827, opened 2020-11-07
Rscircus master patch 21650

rscircus:


rscircus:
Sorry, my mistake. Just created a code mirror which is pushing automatically from GitLab (because Web IDE) and got confused as it is 00:55 in the night. :sleeping: ... actually, I got to :sleeping_bed: now (n8).

PR #837, opened 2020-11-26
Add an option to save data to custom server (csv, txt...)

misko903:
Hello,
great project, I like it, but i am missing the function to simple save data to my Raspbery Pi. I am using WeeWX to generate weather reports and this would be a great addon to the graphs in my project.

Thanks,
Michal,
Slovakia


ricki-z:
@misko903 could you please add such a feature request as an issue (and not a comment to a pull request)? This is completely wrong here.
And: this feature exists in this firmware. Just look for "Send data to custom API". There you can configure a server and a path to a script on i.e. a Raspberry PI. If configured the firmware will send a JSON object to this address. There is a sample script at https://github.com/opendata-stuttgart/madavi-api/blob/master/data_simple.php that will take this object and save the data to a CSV file.

PR #839, opened 2020-11-29
Specify platformio external library dependencies as author/libname

bertrik:
This exactly specifies the library and is much clearer than just a library number.

So for exampe, instead of library "576" it now specifies "marcoschwartz/LiquidCrystal_I2C" and the comment can be removed.


bertrik:
BTW, you can verify I picked the correct libraries with author and library name by typing
  pio lib info 576
for example, and see that the name is indeed the intended library

PR #845, opened 2020-12-06
First step to add HECA dehumidifier

DeeKey:


DeeKey:
This is work in progress. Not working yet!

PR #847, opened 2020-12-09
Battery monitor feature

gvidinski:
<p>I would like to introduce to you a small feature as an addition to the main functionality. I called it “Battery monitor”. It allows you to monitor the state of the battery if UPS topology is used to power the board by showing current battery capacity percentage and its voltage which might be useful for many advanced users I believe.</p>
<p>The feature can be enabled through the configuration menu. When enabled user will have to enter few parameters – threshold voltage for both, the discharged [Battery U min (mV)] and the fully charged [Battery U max (mV)] state along with the maximum input voltage at the voltage divider [ADC divider U max]. In case when NodeMCU<sup>1)</sup> is used and battery voltage is greater than 3.3V user will have to add additional resistor (R<sub>1</sub>) to the monitor line, in series with the positive terminal and the A0 input of the board (Figure 1):</p>
<p>Figure 1. </p>

<p>The value of the R<sub>1</sub> can be found by solving the following equation (1):</p>

(1),
<p>where R<sub>2</sub> = 220kΩ and R<sub>3</sub> = 100kΩ both forming the internal voltage divider provided inside the NodeMCU<sup>1)</sup> board; V<sub>in</sub> is the voltage at the battery’s positive terminal; V<sub>out</sub> is the maximum input voltage of the ESP8266’s 10-bits ADC which is always 1V. Internal voltage divider provides sufficient solution for input voltages ≤ 3.3V without having the need of R<sub>1</sub> to be installed. But having for instance a Li-Polymer single cell battery we will need to be able to read voltage of at least 4.2V. For example solving equation (1) for Vin = 5V will give us a standard value for the R<sub>1</sub> resistor of 180kΩ.</p>
<p>Of course there could be another approach especially if we do not have a resistor with the exact value of 180kΩ for R<sub>1</sub> on hand but we have one with value of 155kΩ instead. By simply using the values of all the resistors in the voltage divider and solving (2):</p>

(2),
<p>where V<sub>out</sub> = 1V; R<sub>1</sub> = 155kΩ; R<sub>2</sub> = 220kΩ; R<sub>3</sub> = 100kΩ will give us the [ADC divider U max] value for the configuration – V<sub>inMAX</sub> = 4.75V. </p>
<p>The code takes into account if we build it for ESP8266 or ESP32 platform. For ESP32 platform I’ve used a definition where pin 36 (GPIO36) is selected for the battery monitor input which corresponds to the ADC1 channel 0 (ADC0) of the ESP32. (By the time I was writing this I didn’t have any available Node32<sup>2</sup> so the definition needs to be tested and eventually changed.) In this case user will have to implement his own voltage divider by using only two resistors as shown in Figure 2:</p>
<p>Figure 2.</p>

<p>Then the following equation (3) have to be used:</p>

(3),
<p>where V<sub>in</sub> is the voltage at the battery positive terminal; V<sub>out</sub> = 3.3V (the maximum meaningful input voltage for the 12-bit ADC’s analog inputs of the ESP32); R<sub>1</sub> & R<sub>2</sub> by your choice. Just pick large enough value in order of hundreds kilo Ohms to keep the current that will be drawn from the battery at insignificant levels. I would suggest using 220kΩ (R<sub>1</sub>) and 330kΩ (<sub>R2</sub>) standard 1% resistors. This will give readings of maximum 5.5V with small enough current draw (7.6 µA @ 4.2V). </p>
<p>In Table 1 are shown different solutions for different input voltage ranges for both platforms using standard resistors of 1% tolerance and will draw current as low as 10µA form the battery.</p>
<p>Table 1.</p>
<table>
<thead>
<tr>
<th>Board</th>
<th align="center" colspan="4">NodeMCU<sup>1)</sup></th>
<th align="center" colspan="4">Node32<sup>2)</sup></th>
</tr>
</thead>
<tbody><tr>
<td>V<sub>inMAX</sub></td>
<td align="center">5V</td>
<td align="center">10V</td>
<td align="center">15.2V</td>
<td align="center">19.2V</td>
<td align="center">5V</td>
<td align="center">10V</td>
<td align="center">15.2V</td>
<td align="center">19.2V</td>
</tr>
<tr>
<td>R<sub>1</sub></td>
<td align="center">180k</td>
<td align="center">680k</td>
<td align="center">1.2M</td>
<td align="center">1.6M</td>
<td align="center">180k</td>
<td align="center">680k</td>
<td align="center">1.2M</td>
<td align="center">1.6M</td>
</tr>
<tr>
<td>R<sub>2</sub></td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">220k</td>
<td align="center">220k</td>
<td align="center">220k</td>
<td align="center">220k</td>
</tr>
<tr>
<td>R<sub>3</sub></td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">-</td>
<td align="center">100k</td>
<td align="center">100k</td>
<td align="center">100k</td>
<td align="center">100k</td>
</tr>
<tr>
<td>Applications</td>
<td align="center"><sup>3)</sup></td>
<td align="center"><sup>4)</sup></td>
<td align="center"><sup>5)</sup></td>
<td align="center"><sup>6)</sup></td>
<td align="center"><sup>3)</sup></td>
<td align="center"><sup>4)</sup></td>
<td align="center"><sup>5)</sup></td>
<td align="center"><sup>6)</sup></td>
</tr>
</tbody></table>
<p>The UI is organized in the following manner. All the settings for the Battery monitor have to be entered form the “Configuration” page, “More settings” tab, under the “Enable battery monitor” checkbox (Figure 3.). By enabling it another three more input fields will become visible. For the Battery discharged, fully charged voltage and for the maximum input voltage at the divider (Figure 4). Then the actual battery state can be obtained at the “Device status” page in the form of “Capacity: [%]([V])” (Figure 5.).</p>
<table>
<tr>
</tr>
<td width="400">Figure 3.</td>
<td width="400">Figure 4.</td>
<tr>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td width="400">Figure 5.</td>
</tr>
<tr>
<td>

</td>
</tr>
</table>
<p>All the measurements are done by averaging 3 consecutive readings taken in periods of 1 second, then followed by 10 seconds pause period. Also no reading will be taken during the report period, when data has to be send to the external web resources (taking into account the “send_now” state). In addition the Battery state data is added to the json result. This way users can track the state of the battery and also may develop customized battery state charts on their platforms.</p>
<p>Build and testing was done using the PlatformIO under the Visual Studio Code IDE following the “Contributing Guidelines”. </p>
<hr>
<p>
<sup>1)</sup> Developments boards based on ESP8266 chips – NoneMCU v2 & v3.<br />
<sup>2)</sup> Development boards based on ESP32 chip – Node32.<br />
<sup>3)</sup> Single cell (1S) Li-Ion or Li-Polymer 3.7V batteries (2.8V ~ 4.2V).<br /> 
<sup>4)</sup> Double cell (2S) Li-Ion or Li-Polymer 7.2V battery packs (5.6V ~ 8.4V).<br />
<sup>5)</sup> Six cell lead acid 12V battery (10.5V ~ 13.8V).<br />
<sup>6)</sup> Four cell (4S) Li-Ion or Li-Polymer 14.4V battery packs (11.2V ~ 16.8V).<br />
</p>


DeeKey:
Great job done!
Small remarks from me:
To my mind the BATTERY_CAPACITY should be renamed to BATTERY_CHARGE. Capacity will slowly degrade over time due to natural reasons. But what you want to show is not capacity, but the charge left in the battery.
Also to simplify the understanding I propose:
const char INTL_BATTERY_U_MIN[] PROGMEM = "Discharged voltage (mV)";
const char INTL_BATTERY_U_MAX[] PROGMEM = "Charged voltage (mV)";
const char INTL_ADC_DIVIDER_U_MAX[] PROGMEM = "ADC divider max voltage (mV)";

glosair:
Would it be useful to turn off the 5V to the SDS011 between measurements as I'm just about to start the next version of my ESP board.

gvidinski:
Thank you @DeeKey for your remarks. You are right. Battery charge is the right definition and I just was not sure. I'll try to update the code in the next few days along with the updated labels proposed by. Listen, can you send me an email in order to get in touch? I would like to discuss a few things with you. Thanks.

git-user44:
"battery_last_value" isn't used anywhere. Why not sum the analogue reading straight into "battery_sum"  and save a few bytes ?

DeeKey:
@gvidinski if it is not something which need private discussion, then better ask question on the official forum. So that others can join discussion and add valuable information:
forum.sensor.community

gvidinski:
_Update_

Hello everyone!
Sorry for the delay. 
I've done some changes as they were suggested by @DeeKey & @git-user44. Still I left without changing a few of them though. I think they need a little bit of discussion. For instance the two of them - 'Discharged voltage (mV)' and 'Charged voltage (mV)' - as they describe to particular properties of the battery in use, to be exact: 1)The maximum voltage when the the battery is fully charged and where the most Over-voltage protection circuits will cut-of charging; 2) The minimum voltage when the battery is considered to be fully discharged and where more protection boards will cut-of any drain form the battery. So on one hand probably it should sound something like "Maximum voltage when fully charged (mV)" and "Minimum voltage when fully discharged (mV)". But will that be too long for those strings where every byte counts? On other hand the meaning of "Charged voltage" and "Discharged voltage" does not exactly explain the two properties. Not to mention that it cannot be translated at all in my native language so the "Charged voltage" should be e.g. "Максимално напрежение при напълно заредена батерия (mV)". Too long, right? My intention using just "Battery U max" & "Battery U min" instead was to keep it simple while it is still meaningful enough. 
I'm just saying my concerns. 
Of course there is possibility for having totally different phrases for every language instance. Some additional help will be welcome.

DeeKey:
Sorry, as per discussion with core developers I had to admit that this pull request will not be accepted upstream.
There is  a constant strive to minimize the code memory print and make it easier for users to add hardware.
In that case it is better to add some i2c hardware sensor then to add such a heavy and hard for ordinal user solution.
Still let it be as a pull request for advanced users and hackers :)

stewecar:
i am sad about considering that feature an hacker advanced user's request, but the project has space limits and i have to cope with that :).
I am on a 700mAh battery powered by a 10W solar panel and trying to let it working 24h per day is a pain..also if i live in Italy.
Cheers!

glosair:
There are several problems with the battery system you are proposing.

Firstly you are providing Vin of 5V that is stepped up from the LiPo cell or stepped down from a LiPo battery, this has loses.

As Vin is a regulated 5V any measurement of Vin as apposed to the actual cell voltage is meaningless.

Secondly the LDO (Low Drop Out) regulator supplying the 3.3V for the ESP8266 is dumping 33% of the chips consumption as heat.

Thirdly you are powering a USB to serial chip constantly draws power.

My suggestion would be to use a LiFo cell and dump the LDO regulator and serial to usb chip, use a small efficient boost converter to power the dust sensor and you are all good.

gvidinski:
Thanks @DeeKey for the answer!

Guys, we can continue at my discussion board in order not to make it too noisy here.
https://github.com/gvidinski/sensors-software/discussions/1

PR #855, opened 2020-12-21
Update intl_nl.h

Laav:
Sensors > Sensoren (https://twitter.com/onzetaal/status/890160690157244416)


bertrik:
Sensors vs sensoren is a bit arbitrary, see also https://www.vandale.nl/gratis-woordenboek/nederlands/betekenis/sensor

PR #862, opened 2021-01-05
Extend pw length to 130 bytes

ainodevries:
Please consider increasing the maximum password length from 65. InfluxDB tokens are at least 88 characters and are cut off silently at the moment.


ricki-z:
This won't work. InfluxDB2 auth tokens and 'username/password' basic auth are working completely different. For InfluxDB2 we need to implement another login mechanism. And please push changes against the beta branch. The master is the latest stable release.

ricki-z:
@ainodevries 
The new beta firmware should have a longer password field. So the compatibility mode should work now. Could you please test this?

PR #870, opened 2021-02-05
Fixes wrong ISO-3166-2 code for Serbian and provides proper translations

marxo:
This commit fixes wrong translation for Serbian strings, changes the script to default Cyrillic (Latin should be added with sr_RS-Latn), and fixes the wrong ISO-3166-2 code `rs` for Serbian language used through the codebase to the proper `sr`.


ricki-z:
We use the 2 letter coutry codes (so we can use 2 letters even for i.e. swiss german, austrian german and german german).
And I have asked the initial translator/commiter if the latin charset is okay. He said most people in Serbia can read this. So could you please discuss this with

marxo:
The official script for Serbian is Cyrillic, and I would be happy to provide both Latin and Cyrillic translations. Also, the quality of the translation was improved by this PR. 

In regards to the ISO-3166-2 code, I think that language codes are standardized through ISO 639 (-1 or -2), so that route might be better to take, given that many dialects will appear in a language, scripts and variants, this will allow for much better handling of langauge variations, such as Serbian (sr) having two scripts (sr_Cyrl, sr_Latn), Chinese (zh) having different dialects (zh_TW, zh_CN) and scripts (zh-Hans, zh-Hant), or language variations spreading through multiple countries like German in Austria (de_AT), Germany, Switzerland and so on.

I also volunteer to do all the work for this transition to happen.

ricki-z:
Just that I can close that tab: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
The number of languages is a matter of maintenance, at every new string we need the translation for all languages. Most countries have an official language that we should focus on.
And using only 2 letters per language saves us some space as we need this on some places in the firmware. We need every single free byte ... (we can only use 700 kByte for the firmware image, otherwise we won't be able to do OTA updates)

ricki-z:
I have merged another pull request today that also updated some strings.

PR #928, opened 2021-11-11
Facilitate translation of "device status" screen.

johnjohndoe:
# Description
+ Extract English texts into constants.
+ Add German translations.

# Draft / To be discussed
Here are a few topics where I need your expertise before I can mark this PR as ready for review.
- [ ] Do **new** constants have to be added to all translation files?
- [ ] Clarify where to use `FPSTR()` or `F()`.
- [x] Clarify what `Last OTA` means.
- [ ] Clarify what `Data Send Return` means.

# Testing
- :warning: The changes have **not** been tested on a device.


ricki-z:
> # Draft / To be discussed
> Here are a few topics where I need your expertise before I can mark this PR as ready for review.

>     * [ ]  Do **new** constants have to be added to all translation files?
Yes, they have to be added to all files. Translators may otherwise forget to translate some of the strings.

>     * [ ]  Clarify where to use `FPSTR()` or `F()`.
I have found the following:

The purpose of the PSTR() macro is to configure the string to be used directly from flash memory. The string is then not loaded to dynamic memory at runtime as it would be without PROGMEM specifier set by the macro.
The purpose of the FPSTR() macro is to cast a string to 'dummy' FlashStringHelper type to help compiler choose the right overloaded function if dynamic memory and flash memory version of a function is available. An example would be the print function.
The F() macro combines this two macros. It makes the string a PROGMEM string and casts it to FlashStringHelper.

>     * [ ]  Clarify what `Last OTA` means.
"Last OTA" is the time of the last lookup for an new firmware version. It's not the time the last OTA was done.

>     * [ ]  Clarify what `Data Send Return` means.
'Data Send Return' is the result code of the last data transmission.

johnjohndoe:
Thank you for your response.

> > * [ ]  Do **new** constants have to be added to all translation files?
>
> Yes, they have to be added to all files. Translators may otherwise forget to translate some of the strings.

Do you want me to put the **English text** in the language files where I am not able to provide the translation myself?

> > * [ ]  Clarify where to use `FPSTR()` or `F()`.
> 
> [...] The F() macro combines this two macros. It makes the string a PROGMEM string and casts it to FlashStringHelper.

That sounds very much like that you want me to use the F() macro everywhere. Okay?

> 
> > * [ ]  Clarify what `Data Send Return` means.
> 
> 'Data Send Return' is the result code of the last data transmission.

Is this the status code of the HTTP response?

PR #934, opened 2021-11-19
added Next PM support

pjgueno:
First version


pjgueno:
Just 1 conflict...

pjgueno:
I found a mistake which bring the default wifi credentials after loading the json config. I correct soon.

PR #947, opened 2022-01-18
Make WiFi maximum TX power configurable

basilfx:
In an effort to save power, make this option configurable. The default is `true`, which equals the current behavior of maximizing TX power.


ricki-z:
It's better do not make this configurable. There will be users changing this without knowing what it's doing. And we will have more support requests.

basilfx:
As an alternative: move this to advanced settings (next to debug level etc.)?

ricki-z:
We could move it to the advanced settings. This page may need some kind of confirmation that users know what they are doing ...

PR #1004, opened 2023-09-27
patch to re-enable the ability to use WiFi with no password, like freifunk routers use to use.

modem-man-gmx:
as pointed by mail some days ago, this is the patch to make the Airrohr able to use Freifunk routers again. I worked in quite early versions, but between 2018 and 2021 the semantics of the espressif API changed (apparently when WPA3 came up), giving an empty password as argument 2 didn't worked anymore with WPA2/3 APs.
Tested with Joy-IT OR750i Router running Freifunk-Hannover Firmware vH34 / gluon-v2022.1.3 which is the same compared to Freifunk-Stuttgart, except the preset routing.

Would highly appreciate to see this in the master soon ;-)

### DIFF
--- a/airrohr-firmware/airrohr-firmware.ino
+++ b/airrohr-firmware/airrohr-firmware.ino
@@ -57,7 +57,7 @@
 #include <pgmspace.h>
 
 // increment on change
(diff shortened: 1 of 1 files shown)


modem-man-gmx:
Thank You!

PR #1041, opened 2024-10-18
Patch/freifunk fix2

modem-man-gmx:
please take this extension to my earlier freifunk fix (handling WiFi hotspot without passwords). From field usage, the earlier patch has shown to not cover the rare use cases: reconnect + switching from WPA AP to Hotspot AP. Fixed now.


ricki-z:
Please push against the beta branch as mentioned in the README.MD.

PR #1048, opened 2025-02-03
enable IIR filtering for BME280

richdove:


Phaze-III:
This patch is running on 3 sensors for a while now and quite effectively smoothens out the temperature and pressure measurements of the BME280. In the graph below (3m interval) sensor-1 (green line) already has the IIR-filter enabled. sensor-2 installed the IIR-enabled firmware around 18:00. sensor-3 runs the original firmware without IIR-filter.

Screenshot 2026-07-12 at 13 57 28

The patch enables maximum filtering (IIR_16). I've seen different recommendations for setting the IIR filter coefficient when used in 'weather monitoring'. Bosch [1] recommends no filter for weather monitoring so you get the raw fluctuating data. On the other hand for example on https://pcbsync.com/bme280-arduino/ it states:
> For weather monitoring, aggressive filtering (FILTER_X16) works well since weather changes slowly.

@richdove : what was your reasoning to use IIR_16?

[1] https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bme280-ds002.pdf page 19

